PR #9, opened 2021-05-05
Save selected folder in SaveExternalFilesActivity

JorisBodin:


app/src/main/java/com/infomaniak/drive/data/models/UISettings.kt
```diff
@@ -33,6 +33,23 @@ class UISettings(val context: Context) {
         }
     }
 
+    fun getSaveExternalFilesPref(): Triple<Int, Int, Int> {
```
Kilian-Perisset, line 36: I would use a Bundle or a data-object instead of a Triple, but anyway it's correct !

Kilian-Perisset (approved):
Some typos to manage, good for me 👍

PR #21, opened 2021-05-19
Fix upload

sirambd:
# Fix: some errors with upload
 - Taked photo and video from kDrive doesn't work
 - Error data file while uploading
 
# News
 - All data to be uploaded are saved in cache temporarily, to ensure the data during the upload
 - Photos and videos taken from kDrive app are also temporarily cached, and deleted once the upload is finished
 
> All temporary files are deleted once their operations are completed

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
@@ -19,6 +19,7 @@ package com.infomaniak.drive.data.models
 
 import android.content.Context
 import android.net.Uri
+import com.infomaniak.drive.data.sync.UploadMigration
 import com.infomaniak.drive.utils.AccountUtils
 import com.infomaniak.drive.utils.RealmModules
 import io.realm.Realm
@@ -39,6 +40,7 @@ open class UploadFile(
     var fileName: String = "",
     var fileSize: Long = 0L,
     var identifier: String = UUID.randomUUID().toString(),
+    var originalLocalUri: String = "",
     var remoteFolder: Int = -1,
     var type: String = Type.SYNC.name,
     var uploadAt: Date? = null,
@@ -55,6 +57,8 @@ open class UploadFile(
         }
     }
 
+    fun isSync() = type == Type.SYNC.toString()
+
     enum class Type {
         SYNC, UPLOAD, SHARED_FILE
     }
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
```diff
@@ -63,7 +69,9 @@ open class UploadFile(
         private const val DB_NAME = "Sync.realm"
         private const val ONE_DAY = 24 * 60 * 60 * 1000
         private var realmConfiguration: RealmConfiguration = RealmConfiguration.Builder().name(DB_NAME)
+            .schemaVersion(1) // Must be bumped when the schema changes
```
Kilian-Perisset, line 72: I think it's better to put the schema version on a higher line in code (like in a constant) to facilitate the understanding of this codeline ?

app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
```diff
@@ -134,9 +142,9 @@ open class UploadFile(
             }
         }
 
-        fun notExists(uri: Uri, lastModified: Date): Boolean {
+        fun canUpload(localUri: Uri, lastModified: Date): Boolean {
             return getRealmInstance().use { realm ->
-                syncFileByUriQuery(realm, uri.toString())
+                realm.where(UploadFile::class.java).equalTo(UploadFile::originalLocalUri.name, localUri.toString())
```
Kilian-Perisset, line 147: Could we use `toSafeString()` method from `Uri` object instead of `toString()` ?

app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
sirambd: I think it's a good idea but here we know we're talking about a version so I don't really see the point, except to make it more complex, because if we need to know the current version, we'll have to look for it, whereas here we can see the version directly 😄

app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
sirambd: No need for that here, we don't do any logging here, no need to format this uri knowing it's just a file path

Kilian-Perisset: I see, thanks for your answer 👍

app/src/main/java/com/infomaniak/drive/utils/Utils.kt
```diff
@@ -322,4 +325,18 @@ object Utils {
             grantResults.isNotEmpty() && grantResults[0] == PackageManager.PERMISSION_GRANTED
         } else false
     }
+
+    fun copyDataToUploadCache(context: Context, uri: Uri, fileModifiedAt: Date): Uri {
```
JorisBodin, line 329: Function "copyDataToUploadCache" is never used

app/src/main/java/com/infomaniak/drive/utils/SyncUtils.kt
```diff
@@ -159,7 +159,7 @@ object SyncUtils {
         }
     }
 
-    private fun Context.cancelContentObserver() {
```
JorisBodin, line 162: Why?
image

JorisBodin (changes requested):

JorisBodin (approved):

PR #26, opened 2021-05-20
Fixed external user bug (manage rights)

Kilian-Perisset:
Before, you were able to set "Manage" rights to an external user in a share.
A toast/snackbar was indicating a bad action -> now we can't perform the permission change

Signed-off-by: Kilian Périsset <kilian.perisset@infomaniak.com>


JorisBodin (changes requested):
Remove isExternalUser

Issue #30, opened 2021-05-21
Deleting last file from webapp creates refresh issue


Kilian-Perisset:
To reproduce :
1) Create a folder with 3 files
2) Delete 2 of them in mobile app
3) Delete the last one in web app
4) Folder is not refreshed, when you refresh : file is still here

Kilian-Perisset:
Fixed by : https://github.com/Infomaniak/android-kDrive/pull/33

PR #33, opened 2021-05-25
Fixed cache refresh after last-file deletion

Kilian-Perisset:
Due to a bad type returned from realm. 
Fixed in collaboration with @sirambd,  thank you 🙏 

Signed-off-by: Kilian Périsset <kilian.perisset@infomaniak.com>


Kilian-Perisset:
It fixes : https://github.com/Infomaniak/android-kDrive/issues/30

JorisBodin (approved):

Issue #24, opened 2021-05-20
Sharing/Permissions reccurence issue


Kilian-Perisset:
When you assign permissions/shares to an existing folder, child files will have difficulties to display same shares than parent.
This bug is not present in web-app.

Issued from these comments :
- https://github.com/Infomaniak/android-kDrive/issues/22#issuecomment-844998767
- https://github.com/Infomaniak/android-kDrive/issues/22#issuecomment-845001668

To reproduce : 
1. Create a folder
2. Create files in it (docs, pictures, etc.)
3. Assign shares in parent folder (step 1)
4. Child will not display shared users neither share informations immediately, you have to refresh... (bug/enhanceable ux)

Kilian-Perisset:
Bug has apparently been fixed.
- For folders, we need to specifiy if we want to inherit from parent folder
- For file it's automatically inherited

but some actions in FileInfoActions are blocked because file `rights` are null on creation. 
I'm trying to determine why..

Kilian-Perisset:
https://github.com/Infomaniak/android-kDrive/pull/35

PR #35, opened 2021-05-25
Fixed rights deletion after update

Kilian-Perisset:
Children rights of a folder were deleted from Realm on parent data fetch.
It fixes this bug.


Kilian-Perisset:
FIxes : https://github.com/Infomaniak/android-kDrive/issues/24

JorisBodin (approved):

PR #50, opened 2021-06-07
Replaced fallback drawable by layout in PreviewPicture

Kilian-Perisset:
+ Removed warning from touch image view


JorisBodin (changes requested):
Do not delete the loading of the thumbnail!

PR #110, opened 2021-07-09
Upload file auto deletion

Kilian-Perisset:
Implemented auto-delete of files when uploaded since more than **one month**

Feature has been tested on Android 11 (R) and 9 (Pie)

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/models/SyncSettings.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/models/SyncSettings.kt
@@ -31,7 +31,8 @@ open class SyncSettings(
     var syncFolder: Int = -1,
     var syncImmediately: Boolean = false,
     var syncInterval: Long = 0,
-    var syncVideo: Boolean = false
+    var syncVideo: Boolean = false,
+    var deleteAfterSync: Boolean = false
 ) : RealmObject() {
 
     fun setIntervalType(intervalType: IntervalType) {
--- a/app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
@@ -100,13 +100,14 @@ open class UploadFile(
             return realm.where(UploadFile::class.java).equalTo(UploadFile::uri.name, uri)
         }
 
-        private fun pendingFilesQuery(realm: Realm, folderID: Int): RealmQuery<UploadFile> {
-            return realm.where(UploadFile::class.java)
-                .equalTo(UploadFile::remoteFolder.name, folderID)
-                .equalTo(UploadFile::userId.name, AccountUtils.currentUserId)
-                .equalTo(UploadFile::driveId.name, AccountUtils.currentDriveId)
-                .isNull(UploadFile::uploadAt.name)
-                .isNull(UploadFile::deletedAt.name)
+        private fun pendingFilesQuery(realm: Realm, folderId: Int? = null): RealmQuery<UploadFile> {
+            return realm.where(UploadFile::class.java).apply {
+                folderId?.let { equalTo(UploadFile::remoteFolder.name, it) }
+                equalTo(UploadFile::userId.name, AccountUtils.currentUserId)
+                equalTo(UploadFile::driveId.name, AccountUtils.currentDriveId)
+                isNull(UploadFile::uploadAt.name)
+                isNull(UploadFile::deletedAt.name)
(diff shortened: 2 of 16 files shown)


supaeasy:
Deletion on the mobile device or in kDrive? If first: great! Even better: could there be an option to adjust that timeframe to days or even hours?

Kilian-Perisset:
Hello @supaeasy !
Thank you for your contribution to the app, these are very appreciative and useful comments 

This feature deletes files from device once uploaded on kDrive, but the current delay is fixed to one month. 
About timeframe, it's currently defined to one month and not editable, but everything is possible :

@alxschilling @JorisBodin What do you think about defining a custom timeframe before auto-deletion of files ?

supaeasy:
Love it! I really missed the auto delete and a month is a very good start. Thank you.

JorisBodin:
@supaeasy the option must be identical on iOS and Android
And in the end we will do as on iOS. Propose to delete the photos every 10 launches of the app or as we have uploaded 50 photos.
To avoid spam or too much time to delete these photos

Kilian-Perisset:
Hi,

So the feature has been harmonized with iOS.
 
@supaeasy FYI, it's no longer based on deletion delay, but on the amount of deletable files.
So : 
- When 50+ files will be able to be deleted
AND
- Every 10 app launches 

it will give you the ability to allow the deletion of your files through a simple dialog. 
So that you're sure of what you delete and you can cancel if you're not sure ;) 

This feature will be implemented ASAP in production

supaeasy:
@Kilian-Perisset 

To be honest I don't see a point in this logic. So when I want to free space on my phone from, lets say a 1 GB Video,  I have to take 49 Photos and open and close the app 10 times? Seems *very* unhandy to me. 

I would at least expect a button "delete all uploaded files now", would prefer an automation setting to "delete everything older than x days that has been uploaded"

supaeasy:
> We know that the deletion dialog logic is clearly enhanceable but there are 3 complications : 
> 1. We need to ensure the file has been correctly synchronized with Drive before deleting it from device (or asking to user)

Wouldn't be a checksum an option to ensure that? 
> 2. For new versions of Android, we use MediaStore deletion request and no longer direct file deletion, so it implies to remove files "in bulk".

Sorry I don't understand what that means. 
> 3. If we only take time as an indicator (and not launches nor uploaded files), the app could ask you everyday if you want to remove some files, and we want to avoid that ;) 

If settings are set to that option it would be unnecessary to ask. That would also be a very annoying dialogue if not mutable via checkbox/button (ok and never ask again). Also I don't see a risk if upload is verified, see point 1. To me, logic would not be "delete pictures taken x days ago" but "delete pictures _uploaded_ x days ago"
> 
> So, today we're looking for a solution to meet the above-mentioned complexities and we'll provide a better balanced idea (based on your review).
> 
> Anyway, your comments are always welcome, and we will always do our best 👍 

@Kilian-Perisset Let me answer directly above. 
I know I might sound picky or demanding.  I really don't mean to be. Great app, great development! But this really is a key feature to me, that's why I am a little pushy here.

Kilian-Perisset:
> > 1. For new versions of Android, we use MediaStore deletion request and no longer direct file deletion, so it implies to remove files "in bulk".
> 
> Sorry I don't understand what that means.

It means that we use a "Library" provided by Android to perform actions on media files. 
So, we don't delete files directly after their upload (like Nextcloud/kDrive 3.x), we wait a certain amount before asking user if he wants to delete specific files through a dialog. 


> > 1. If we only take time as an indicator (and not launches nor uploaded files), the app could ask you everyday if you want to remove some files, and we want to avoid that ;)
> 
> If settings are set to that option it would be unnecessary to ask. That would also be a very annoying dialogue if not mutable via checkbox/button (ok and never ask again). Also I don't see a risk if upload is verified, see point 1. To me, logic would not be "delete pictures taken x days ago" but "delete pictures _uploaded_ x days ago"
> 

Currently we don't implement checksum, we would need to develop it in backend-side. It's hard to determine if a file is actually uploaded and complete on server-side. 
So, we prefer to let the user decide or not if he wants to remove his files (rather than delete it without guarantee, it's a dangerous option .. it would be unappreciated if an option like this cause data loss ). 
And later, once checksum feature implemented, we'll be able to provide a more complete option 👍 

----

A new version has been pushed and at every launch, if there are 10+ files uploaded, it'll ask you if you want to remove them.

image

image

JorisBodin (approved):

PR #132, opened 2021-07-19
limit get recents from home to 1 page

sirambd:


app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -350,7 +350,11 @@ class MainViewModel(appContext: Application) : AndroidViewModel(appContext) {
         }
     }
 
-    fun getRecentChanges(driveId: Int, forceDownload: Boolean = false): LiveData<FileListFragment.FolderFilesResult?> {
+    fun getRecentChanges(
```
Kilian-Perisset, line 353: on one line, is it possible ?

Kilian-Perisset (approved):

PR #200, opened 2021-08-23
fix characters that may crash download manager

sirambd:


Kilian-Perisset (commented):
Is it the only place we need to "normalize" file names ?

sirambd:
> Is it the only place we need to "normalize" file names ?

@Kilian-Perisset yes for the downloadManager

Kilian-Perisset (approved):

PR #230, opened 2021-09-06
delete only if realm object is valid

sirambd:


Kilian-Perisset (changes requested):
I think it's better to directly check if file is valid in `updateFile` method before callback, what do you think ?

sirambd:
> I think it's better to directly check if file is valid in `updateFile` method before callback, what do you think ?

Yes i agree, it's a good idea

Kilian-Perisset (approved):

PR #236, opened 2021-09-07
Don't start retry or start upload when the mobile storage is low

sirambd:
Warn the system not to restart the sync when the device storage is low.


Kilian-Perisset:
@sirambd Do you have some news ? 
Is it possible to replace this check by "checkIfMemoryIsLow" (we don't really care of storage in case of an upload)

PR #271, opened 2021-09-22
Retry on api upload error

sirambd:


app/src/main/java/com/infomaniak/drive/data/api/UploadTask.kt
```diff
@@ -67,6 +67,7 @@ class UploadTask(
         uploadNotification = context.uploadProgressNotification()
         uploadNotification?.apply {
             setContentTitle(uploadFile.fileName)
+            setProgress(100, 0, true)
```
JorisBodin, line 70: Duplication

JorisBodin (changes requested):

JorisBodin (approved):

app/src/main/java/com/infomaniak/drive/data/api/UploadTask.kt
sirambd: Fixed thank you

PR #307, opened 2021-10-20
Fix realm selected file

JorisBodin:


app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
```diff
@@ -71,6 +71,8 @@ open class FileAdapter(
 
     fun getFiles() = fileList
 
+    fun getValidItemSelected() = if (itemSelected.isManaged) itemSelected.filter { it.isValid } else itemSelected
```
KevinBoulongne, line 74: Can you add the return type to this function ?
It would help to understand what we are talking about here.

PR #343, opened 2021-11-10
Fix realm live

JorisBodin:


app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
```diff
@@ -353,9 +357,7 @@ open class FileAdapter(
     }
 
     private fun isSelectedFile(file: File): Boolean {
-        return itemsSelected.find {
-            (it.isManagedAndValidByRealm() || it.isNotManagedByRealm()) && it.id == file.id
-        } != null
+        return itemsSelected.find { it.isUsable() && it.id == file.id } != null
```
Kilian-Perisset, line 360: Prefer `.any { it.isUsable() && it.id == file.id }` instead of find != null :smirk:

KevinBoulongne (approved):

sirambd (approved):

PR #377, opened 2021-11-26
Categories #01: Added Models

KevinBoulongne:
Signed-off-by: Kevin Boulongne <kevin.boulongne@infomaniak.com>

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
@@ -19,8 +19,11 @@ package com.infomaniak.drive.data.cache
 
 import com.infomaniak.drive.data.models.DriveUser
 import com.infomaniak.drive.data.models.Team
+import com.infomaniak.drive.data.models.drive.Category
+import com.infomaniak.drive.data.models.drive.CategoryRights
 import com.infomaniak.drive.data.models.drive.Drive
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
```diff
@@ -129,4 +129,34 @@ object DriveInfosController {
 
         return teamList.filter { drive.teams.account.contains(it.id) }
     }
+
+    fun getCategories(ids: Array<Int>): List<Category> {
+
+        if (ids.isEmpty()) return emptyList()
+
+        val categories = getRealmInstance().use { realm ->
+            val drive = realm.where(Drive::class.java)
```
sirambd, line 138: Rename `drive` to `currentDrive`, i think it would be better

app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
```diff
@@ -129,4 +129,34 @@ object DriveInfosController {
...
+                .equalTo(Drive::id.name, AccountUtils.currentDriveId)
+                .findFirst()
+
+            drive?.categories?.let {
```
JorisBodin, line 142: Use AccountUtils.getCurrentDrive()

JorisBodin (changes requested):

PR #380, opened 2021-11-26
Categories #04: Management in FileInfo

KevinBoulongne:
Signed-off-by: Kevin Boulongne kevin.boulongne@infomaniak.com


app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
```diff
@@ -141,16 +141,20 @@ object DriveInfosController {
 
             drive?.categories?.let {
                 val categories = it.where()
-                    .`in`(Category::id.name, ids)
```
JorisBodin, line 144: val categories = it.where()
                    .apply { fileCategoriesIds?.let { `in`(Category::id.name, fileCategoriesIds) } }
                    .findAll()

JorisBodin (changes requested):

PR #432, opened 2022-01-13
Refactor SearchFragment + Added SearchViewModel

KevinBoulongne:
Signed-off-by: Kevin Boulongne <kevin.boulongne@infomaniak.com>

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/fileList/SearchFragment.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/fileList/SearchFragment.kt
@@ -34,15 +34,15 @@ import com.infomaniak.drive.data.models.*
 import com.infomaniak.drive.data.models.File.*
 import com.infomaniak.drive.data.models.SearchFilter.*
 import com.infomaniak.drive.ui.fileList.FileListViewModel.*
+import com.infomaniak.drive.ui.fileList.SearchViewModel.*
 import com.infomaniak.drive.utils.*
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/drive/ui/fileList/SearchFragment.kt
```diff
@@ -191,23 +192,16 @@ class SearchFragment : FileListFragment() {
     private fun setToolbarListener() {
         toolbar.setOnMenuItemClickListener { menuItem ->
             if (menuItem.itemId == R.id.selectFilters) {
```
sirambd, line 194: 🤔 Otherwise we can also do like this 😅
```kotlin
            (menuItem.itemId == R.id.selectFilters).also { 
                if (it) safeNavigate(R.id.searchFiltersFragment)
            }
```

sirambd:
> I don't know if this is normal but I have a bug. When I do a search, then I go into the filter view without choosing one and then I go back to my results everything disappears. A message of no results is displayed, but when I refresh everything comes back

I had the same behavior after I opened a file and then made a return.

KevinBoulongne:
Same here, I'll look into that. Thanks!

KevinBoulongne:
> This bug was already existing. I'll try to fix it in this PR.

Fixed.

JorisBodin (approved):

JorisBodin (approved):

sirambd (approved):

PR #708, opened 2022-03-28
Fix MultiSelect download outside of default FileList

KevinBoulongne:
Signed-off-by: Kevin Boulongne [kevin.boulongne@infomaniak.com](mailto:kevin.boulongne@infomaniak.com)

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/fileList/multiSelect/MultiSelectActionsBottomSheetDialog.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/fileList/multiSelect/MultiSelectActionsBottomSheetDialog.kt
@@ -59,7 +59,7 @@ abstract class MultiSelectActionsBottomSheetDialog(private val matomoCategory: S
         configureColoredFolder(areIndividualActionsVisible)
         configureAddFavorites(areIndividualActionsVisible)
         configureAvailableOffline()
-        configureDownloadFile()
+        configureDownload()
         configureDuplicateFile()
         configureRestoreFileIn()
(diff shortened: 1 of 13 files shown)


JorisBodin (approved):

app/src/main/java/com/infomaniak/drive/ui/fileList/multiSelect/MultiSelectActionsBottomSheetDialog.kt
```diff
@@ -157,13 +157,14 @@ abstract class MultiSelectActionsBottomSheetDialog(private val matomoCategory: S
         deletePermanently.isGone = true
     }
 
-    private fun downloadFileArchive() {
-
-        fun downloadArchive(fileIds: IntArray) = liveData(Dispatchers.IO) {
-            emit(ApiRepository.getUUIDArchiveFiles(AccountUtils.currentDriveId, fileIds))
-        }
+    private fun download() {
+        if (navigationArgs.areAllFromTheSameFolder) downloadArchive() else downloadFiles()
```
sirambd, line 161: The problem here is that if you download only one file, it will always download an archive. 
Even if it's still weird this action but it would be nice if it was avoided 😅

sirambd, line 161: The problem here is that if you download only one file, it will always download an archive. 
Even if it's still weird this action but it would be nice if it was avoided 😅

sirambd (changes requested):

sirambd (approved):

PR #718, opened 2022-03-30
Fix pending files doesn't disappear - issue #713

### LINKED ISSUE #713: "import in progress" that does not disappear
Screenshot_1648557230
Screenshot_1648557236

sirambd:
Close #713


JorisBodin (changes requested):
Does not solve the problem of the file not being displayed

JorisBodin (approved):

PR #725, opened 2022-03-31
fix pop stack in preview

FabianDevel:


app/src/main/java/com/infomaniak/drive/views/FileInfoActionsView.kt
```diff
@@ -215,7 +215,9 @@ class FileInfoActionsView @JvmOverloads constructor(
                     availableOfflineSwitch.isChecked = false
                     showSnackBarInvalidFileName(currentFile.name)
                 }
-                findNavController().popBackStack()
+                with(findNavController()) {
+                    if (currentBackStackEntry?.destination?.label == "FileInfoActionsBottomSheetDialog") popBackStack()
```
KevinBoulongne, line 219: Can you put this `FileInfoActionsBottomSheetDialog` in the `strings.xml`, as a non-translatable one?

And use it too in the `main_navigation.xml`.

FabianDevel: done

app/src/main/java/com/infomaniak/drive/views/FileInfoActionsView.kt
```diff
@@ -215,7 +215,11 @@ class FileInfoActionsView @JvmOverloads constructor(
                     availableOfflineSwitch.isChecked = false
                     showSnackBarInvalidFileName(currentFile.name)
                 }
-                findNavController().popBackStack()
+                with(findNavController()) {
+                    if (currentBackStackEntry?.destination?.label == getString(R.string.fileActionBottomSheetDialog)) {
```
JorisBodin, line 219: Use `currentBackStackEntry?.destination?.id`

JorisBodin (changes requested):

app/src/main/java/com/infomaniak/drive/views/FileInfoActionsView.kt
FabianDevel: done

JorisBodin (approved):

PR #787, opened 2022-05-23
Removed the message about creating a folder in select folder fragment

LunarX:
It doesn't mention creating a new folder by using the "+" button anymore


sirambd (approved):

JorisBodin (approved):

JorisBodin:
It does not fix completely https://github.com/Infomaniak/android-kDrive/issues/744
Because if you go to an empty folder in favorites for example ... The message is not good

PR #788, opened 2022-05-24
Automatically selects the filename without the extension

LunarX:
Instead of a cross, the filename (without the extension) is automatically selected when the text field is focused

Closes: #778 
Closes: #752

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
@@ -24,7 +24,6 @@ import android.net.Uri
 import android.os.Bundle
 import android.os.Parcelable
 import android.provider.OpenableColumns
-import android.text.Selection.setSelection
 import androidx.activity.result.contract.ActivityResultContracts.StartActivityForResult
 import androidx.activity.viewModels
 import androidx.core.content.ContextCompat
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
```diff
@@ -91,6 +92,18 @@ class SaveExternalFilesActivity : BaseActivity() {
...
+
+    private fun TextInputEditText.selectAllButFileExtension() {
+        setOnFocusChangeListener { _, hasFocus ->
+            if (hasFocus) {
+                val fileName = (text ?: "").toString()
+                val endIndex = fileName.substringBeforeLast(".").length
+                post { setSelection(0, endIndex) }
+            }
+        }
     }
```
sirambd, line 107: You don't have to do it every time you focus, just once when you create.

sirambd (changes requested):
Waiting for request changes

app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
```diff
@@ -91,6 +93,14 @@ class SaveExternalFilesActivity : BaseActivity() {
...
         fetchFolder()
         setupSaveButton()
+
+        fileNameEdit.selectAllButFileExtension()
+    }
+
+    private fun TextInputEditText.selectAllButFileExtension() {
+        val fileName = (text ?: "").toString()
+        val endIndex = fileName.substringBeforeLast(".").length
+        post { setSelection(0, endIndex) }
```
JorisBodin, line 103: It is necessary to merge the logic with this method
https://github.com/Infomaniak/android-kDrive/blob/master/app/src/main/java/com/infomaniak/drive/data/models/File.kt#L201

JorisBodin (changes requested):

KevinBoulongne (approved):

PR #794, opened 2022-05-25
Folder sharing link can now be shared instead of copied

LunarX:
Closes: #769

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/views/FileInfoActionsView.kt
+++ b/app/src/main/java/com/infomaniak/drive/views/FileInfoActionsView.kt
@@ -274,13 +274,13 @@ class FileInfoActionsView @JvmOverloads constructor(
         }
     }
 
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/infomaniak/drive/views/FileInfoActionsView.kt
```diff
@@ -443,7 +435,7 @@ class FileInfoActionsView @JvmOverloads constructor(
 
         fun addFavoritesClicked() = trackActionEvent("favorite", (!currentFile.isFavorite).toFloat())
         fun colorFolderClicked(color: String) = application?.trackEvent("colorFolder", TrackerAction.CLICK, "switch")
-        fun copyPublicLink() = trackActionEvent("copyShareLink")
+        fun sharePublicLink() = trackActionEvent("copyShareLink")
```
sirambd, line 438: I think that it'll be better to change the track event name

sirambd (changes requested):
Waiting for request changes

PR #815, opened 2022-06-22
Fix move button from pictures fragment with api v2

sirambd:


app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
```diff
@@ -66,7 +66,10 @@ class SelectFolderActivity : BaseActivity() {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_select_folder)
         setSaveButton(customArgs)
-        disabledFolderId?.let { initiateNavigationToCurrentFolder(it, currentUserDrive) } ?: Unit
+
+        disabledFolderId?.let { folderId ->
+            FileController.getFileProxyById(folderId)?.let { initiateNavigationToCurrentFolder(folderId, currentUserDrive) }
```
KevinBoulongne, line 71: Why do you get this file if you don't do anything with it?

Just to check if it still exists?

sirambd: Yes of course

JorisBodin (approved):

PR #827, opened 2022-06-28
Fix upload worker token when app is in background

sirambd:


app/src/main/java/com/infomaniak/drive/data/api/UploadTask.kt
```diff
@@ -123,6 +125,7 @@ class UploadTask(
     private suspend fun launchTask(coroutineScope: CoroutineScope) = withContext(Dispatchers.IO) {
         val uri = uploadFile.getUriObject()
         val fileInputStream = context.contentResolver.openInputStream(uploadFile.getOriginalUri(context))
+        val okHttpClient = KDriveHttpClient.getHttpClient(uploadFile.userId, 120)
```
KevinBoulongne, line 128:
```suggestion
        val okHttpClient = KDriveHttpClient.getHttpClient(userId = uploadFile.userId, timeout = 120)
```

KevinBoulongne (approved):

PR #828, opened 2022-06-29
Fix refresh activities after an upload

sirambd:
**Fix**
- Sometimes when uploading files, the `FileList` was not updated.
> now, the `FileList` is well updated with respect to the upload


app/src/main/java/com/infomaniak/drive/ui/fileList/FileListFragment.kt
```diff
@@ -234,7 +234,10 @@ open class FileListFragment : MultiSelectFragment(MATOMO_CATEGORY), SwipeRefresh
         }
 
         requireContext().trackUploadWorkerSucceeded().observe(viewLifecycleOwner) {
-            if (!isDownloading || isUploading) activitiesRefreshTimer.start()
+            if (!isDownloading || isUploading) {
+                if (isUploading) retryLoadingActivities = true
+                activitiesRefreshTimer.start()
+            }
```
KevinBoulongne, line 240:
```suggestion
            if (isUploading) retryLoadingActivities = true
            if (!isDownloading || isUploading) activitiesRefreshTimer.start()
```

PR #838, opened 2022-07-13
opt out of scoped storage on android Q

sirambd:
Fix: #745


app/src/main/AndroidManifest.xml
```diff
@@ -62,6 +62,7 @@
         android:allowBackup="false"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
+        android:requestLegacyExternalStorage="true"
```
JorisBodin, line 65: Not sure if google play accepts this

PR #848, opened 2022-07-22
Add Scan document

JorisBodin:


app/src/main/java/com/infomaniak/drive/ApplicationMain.kt
```diff
@@ -103,6 +107,15 @@ class ApplicationMain : Application(), ImageLoaderFactory {
 
         runBlocking { initRealm() }
 
+        geniusScanIsReady = try {
+            GeniusScanSDK.init(this, BuildConfig.GENIUS_SCAN_KEY)
+            true
+        } catch (licenseException: LicenseException) {
+            licenseException.printStackTrace()
+            Log.e("GeniusScanSDK", "The license is expired or invalid")
```
sirambd, line 115: I think we should also notify sentry, it can always be useful

sirambd (approved):

PR #865, opened 2022-08-08
Multiples fix

JorisBodin:


app/src/main/res/layout/fragment_home.xml
```diff
@@ -161,7 +161,8 @@
             android:id="@+id/homeViewPager"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
-            android:nestedScrollingEnabled="true" />
+            android:nestedScrollingEnabled="true"
+            android:saveEnabled="true" />
```
sirambd, line 165: It is already false by default no ?

PR #896, opened 2022-09-16
Save fileDetails in DownloadWorker

JorisBodin:


app/src/main/java/com/infomaniak/drive/data/services/DownloadWorker.kt
```diff
@@ -162,9 +162,12 @@ class DownloadWorker(context: Context, workerParams: WorkerParameters) : Corouti
 
     private fun getFileFromRemote() {
         val fileDetails = ApiRepository.getFileDetails(File(id = fileId, driveId = userDrive.driveId))
-        val data = fileDetails.data
-        file = if (data != null) {
-            data
+        val remoteFile = fileDetails.data
+        file = if (fileDetails.isSuccess() && remoteFile != null) {
```
sirambd, line 166: Not necessarily useful to check the api success case, if the `data` exists it is necessarily success

JorisBodin: prevention is better than cure

PR #929, opened 2022-11-16
Change actions visibility to be iso with iOS and webApp

FabianDevel:
Hide color Folder and add Favorite in the FileInfoActionsBottomsheet for external files as it is already done in iOS and web, because coloring an external folder is not possible and favorite external file are not retrieved in FavoritesFragment


JorisBodin (changes requested):
It must be the api that returns this.
No blocking on the front

PR #941, opened 2022-11-22
remove useless api call to activities when refreshing folder

FabianDevel:


app/src/main/java/com/infomaniak/drive/ui/fileList/FileListFragment.kt
```diff
@@ -749,7 +749,7 @@ open class FileListFragment : MultiSelectFragment(MATOMO_CATEGORY), SwipeRefresh
 
                     fileAdapter.isComplete = result.isComplete
                     fileAdapter.hideLoading()
-                    refreshActivities()
+                    if (!binding.swipeRefreshLayout.isRefreshing) refreshActivities()
```
JorisBodin, line 752: Very bad thing to rely on the `swipeRefreshLayout` to do this. Because you could very well load from realm WITH the `swipeRefreshLayout` enabled and then you would lose the activities.

JorisBodin (changes requested):

JorisBodin:
This call is not useless, but necessary.

PR #960, opened 2022-12-21
Fix: file moved stays in favorites file list

### LINKED ISSUE #784: Moving files from the list of favorites makes them disappear from the display
When moving files from your list of favorites, they disappear from the list even though they're still in your favorites. They come back if you manually refresh the list.

They should not be removed form the displayed list.

This also apply to the list of last modifications

FabianDevel:
closes #784

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
@@ -275,9 +275,11 @@ object FileController {
         }
     }
 
-    private fun addChild(realm: Realm, localFolder: File, file: File) {
-        if (localFolder.children.find { it.id == file.id } == null) {
-            realm.executeTransaction { localFolder.children.add(file) }
+    fun addChild(localFolderId: Int, newFile: File, realm: Realm) {
+        getFileById(realm, localFolderId)?.let { localFolder ->
+            if (!localFolder.children.contains(newFile)) {
+                realm.executeTransaction { localFolder.children.add(newFile) }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -872,6 +865,28 @@ object FileController {
...
+                    localFolder.id != file?.parentId && localFolder.id == currentFolder.id -> {
+                        updateFile(localFolder.id, realm) { it.children.remove(file) }
+                    }
+                    else -> Unit
+                }
+            }
+            returnResponse[fileId] = this
+        }
+    }
+
```
sirambd, line 889: You can keep it as before and just simplify it in case you don't want to delete it

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -822,15 +822,8 @@ object FileController {
...
-            FileActivityType.FILE_TRASH -> {
-                if (returnResponse[fileId] == null || returnResponse[fileId]?.createdAt?.time == createdAt.time) { // Api fix
-                    getParentFile(fileId = fileId, realm = realm)?.let { parent ->
-                        if (parent.id == currentFolder.id) removeFile(fileId, customRealm = realm, recursive = false)
-                    }
-                    returnResponse[fileId] = this
-                }
-            }
+            FileActivityType.FILE_TRASH -> removeFileWhenEnteringFolder(true, realm, returnResponse, currentFolder)
+            FileActivityType.FILE_MOVE_OUT -> removeFileWhenEnteringFolder(false, realm, returnResponse, currentFolder)
```
sirambd, line 826: The name is too big, and I think we can largely simplify it, here, for example a name like `handleRemoveAction` is more simple

sirambd (changes requested):

sirambd (approved):

JorisBodin (approved):

PR #1070, opened 2023-10-20
Replace kotlin extension with view bindings in GalleryFragment

LunarX:


app/src/main/java/com/infomaniak/drive/ui/menu/GalleryFragment.kt
```diff
@@ -195,7 +203,7 @@ class GalleryFragment : MultiSelectFragment(MATOMO_CATEGORY), NoItemsLayoutView.
             }
         }
 
-        galleryRecyclerView?.layoutManager = gridLayoutManager
+        binding.galleryRecyclerView?.layoutManager = gridLayoutManager
```
FabianDevel, line 206: As this was nullable, isn't there any chance that this could result in an NPE ? 
It would be safer to keep the nullability for the binding here
Either way, this null check here is useless with "binding"

app/src/main/java/com/infomaniak/drive/ui/menu/GalleryFragment.kt
```diff
@@ -296,7 +304,7 @@ class GalleryFragment : MultiSelectFragment(MATOMO_CATEGORY), NoItemsLayoutView.
     }
 
     fun setScrollbarTrackOffset(offset: Int) {
-        galleryFastScroller?.trackMarginEnd = offset
+        _binding?.let { it.galleryFastScroller.trackMarginEnd = offset }
```
FabianDevel, line 307:
```suggestion
        _binding?.galleryFastScroller?.trackMarginEnd = offset
```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/menu/GalleryFragment.kt
LunarX: It looks like removing the null check should be fine because it's only called inside `onViewCreated` and `onConfiguraitonChanged` but looking back at where this `?`came from, it looks like this was introduced to fix something. Without much knowledge as to why this got here, we'll keep it just to be sure

FabianDevel (approved):

PR #1071, opened 2023-10-20
Replace kotlinx extentsion by viewBinding for dropbox

FabianDevel:


app/src/main/java/com/infomaniak/drive/ui/fileList/ManageDropboxFragment.kt
```diff
@@ -79,7 +79,7 @@ open class ManageDropboxFragment : Fragment() {
 
         fileShareCollapsingToolbarLayout.title = getString(R.string.manageDropboxTitle, navigationArgs.fileName)
 
-        shareLinkContainer.apply {
+        shareLinkContainer.binding.apply {
```
LunarX, line 82: Usually we create functions or expose variables to interact with custom views' internal views. What do we want to do here?

PR #1076, opened 2023-10-23
Replace kotlin extension with view bindings in HomeActivitiesFragment

LunarX:


app/src/main/java/com/infomaniak/drive/ui/home/HomeActivitiesFragment.kt
```diff
@@ -98,7 +107,9 @@ class HomeActivitiesFragment : Fragment() {
     }
 
     fun getLastActivities(driveId: Int, forceDownload: Boolean = false) {
-        (homeTabsRecyclerView?.adapter as? LastActivitiesAdapter)?.apply {
+        if (_binding == null) return
+
+        (binding.homeTabsRecyclerView.adapter as? LastActivitiesAdapter)?.apply {
```
FabianDevel, line 112:
```suggestion
(_binding?.homeTabsRecyclerView?.adapter as? LastActivitiesAdapter)?.apply {
```

FabianDevel (approved):

app/src/main/java/com/infomaniak/drive/ui/home/HomeActivitiesFragment.kt
LunarX: For better readability I would rather go with the `if (_binding == null) return` version

PR #1091, opened 2023-10-25
Replace kotlin extension with view bindings in HomeFragment

LunarX:


app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -451,7 +450,9 @@ class MainActivity : BaseActivity() {
         return bitmap
     }
 
-    fun getMainFab(): FloatingActionButton = binding.mainFab
+    fun getMainFab() = binding.mainFab
+
+    fun getBottomNavigation() = binding.bottomNavigation
```
FabianDevel, line 455: I just thought of it, but it could be val with getter, that woul be smoother to use

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
LunarX: It would indeed. 

Wouldn't keeping it "harder" to use be better though? Like, I don't feel so good letting the Fragment rely on MainActivity so maybe keeping them as functions would prevent people from using them without thinking. That's kinda what I was going for with this

JorisBodin (approved):

PR #1102, opened 2023-10-26
Apiv3-7 home activities

sirambd:
Depends on #1072

PR #1110, opened 2023-10-27
Apiv3-8 File activities

sirambd:
Depends on #1102

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
@@ -785,23 +785,23 @@ object FileController {
         return files?.let { ArrayList(it) } ?: arrayListOf()
     }
 
-    fun getFolderActivities(folder: File, page: Int, userDrive: UserDrive? = null): Map<out Int, FileActivity> {
+    fun getFolderActivities(folder: File, userDrive: UserDrive? = null): Map<out Int, FileActivity> {
         return getRealmInstance(userDrive).use { realm ->
-            getFolderActivitiesRec(realm, folder, page, userDrive)
+            getFolderActivitiesRec(realm, folder, userDrive)
         }
     }
 
     private fun getFolderActivitiesRec(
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -656,7 +656,7 @@ object FileController {
                 result = realm.downloadAndSaveFiles(
                     localFolderProxy = folderProxy,
                     order = order,
-                    loadNextPage = loadNextPage,
+                    isFirstPage = !loadNextPage,
```
FabianDevel, line 659: Won't this cause problems ? because you didn't inverted the check in if after that

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/home/HomeViewModel.kt
```diff
@@ -170,7 +168,7 @@ class HomeViewModel : ViewModel() {
                 previousActivity.action == currentActivity.action &&
                 previousActivity.fileId == currentActivity.fileId
     }
-
+
```
KevinBoulongne, line 171:
```suggestion

```

app/src/main/java/com/infomaniak/drive/ui/fileList/fileDetails/FileDetailsViewModel.kt
```diff
@@ -75,24 +74,24 @@ class FileDetailsViewModel : ViewModel() {
...
                             emit(this)
-                            recursive(page + 1)
+                            recursive(this.cursor)
                         }
                     }
                 }
             }
         }
-        recursive(1)
+        recursive(null)
```
KevinBoulongne, line 95:
```suggestion
        recursive(cursor = null)
```

KevinBoulongne (changes requested):

tevincent (approved):

PR #1111, opened 2023-10-30
Replace kotlinx Extensions by ViewBinding in FileAdapter and FileItemUtils

FabianDevel:


app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
```diff
@@ -241,50 +249,104 @@ open class FileAdapter(
...
             }
+        } else {
+            fileCardView.setCorners(position, itemCount)
+        }
+
+        when (viewType) {
+            DisplayType.LIST.layout -> (binding as CardviewFileListBinding).itemViewFile.setFileItem(file, isGrid)
+            DisplayType.GRID.layout -> (binding as CardviewFileGridBinding).setFileItem(file, isGrid)
+            DisplayType.GRID_FOLDER.layout -> (binding as CardviewFolderGridBinding).setFileItem(file, isGrid)
+        }
```
LunarX, line 346: Would making `setFileItem` an extension of FileItemViews be possible? We'd need some more views inside the data class but we could avoid this `when` which feels like the goal of this "common views interface" through which we interact

app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
```diff
@@ -443,6 +503,17 @@ open class FileAdapter(
         LIST(R.layout.cardview_file_list)
     }
 
+    data class FileItemViews(
```
LunarX, line 506: Would adding the keyword "common" to qualify the views make sens in the name of the class? Do you have an opinion on the subject? Like, to clearly state that these are the views that are common to all three layouts. All of these views are accessible at all time by simply accessing the binding but the ones inside the data class are there for when we want to do something no matter what layout we're dealing with

app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
FabianDevel: Doing this would need to expose the interface publicly, which I wanted to avoid, and thus instantiate it in other places. 
Furthermore I like that the differentiated setFileItem allows a "strong typed"  way for the function, as we sure to understand what code if for what binding. Otherwise I would need to pass 3 or 4 more nullable views in the data class which would mitigate its lisibility and need
So finally this when is a lot clearer than the other solution

app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
FabianDevel: Well this is technically not the case, fileDate and StopUploadButton aren't shared with all bindings. But removing it would make the need of different binding cast inside this adapter, which complexify the code

FabianDevel: However, we could extends the when that you're refering to in your previous comment to avoid other cast, to the cost of a bit more logic duplication. I though it was clearer like this, to keep the when to its simplest form.
Another solution would be to hard cast without verification for exemple fileDate, as it exists only for FileListItemBInding, but I find this less reliable and more error prone in the future

app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
LunarX: Yeah no this is indeed used in many places, I think you've got the best solution

app/src/main/java/com/infomaniak/drive/ui/fileList/FileAdapter.kt
LunarX: We definitely want to keep the nullable views inside FileItemViews. But I don't think it goes against naming it "common" nonetheless

LunarX (approved):

KevinBoulongne (approved):

sirambd (changes requested):
Need to update the logic

JorisBodin (approved):

PR #1112, opened 2023-10-30
Apiv3-9 fix duplication data in gallery

sirambd:
This problem currently exists in production, but I prefer to fix it with v3 following all the changes.
Depends on #1110

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/menu/GalleryViewModel.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/menu/GalleryViewModel.kt
@@ -35,9 +35,9 @@ class GalleryViewModel : ViewModel() {
 
     private var currentCursor: String? = null
 
-    val galleryApiResult = MutableLiveData<Pair<ArrayList<File>, IsComplete>?>()
     private var lastGalleryFiles = arrayListOf<File>()
 
+    val galleryApiResult = MutableLiveData<Pair<ArrayList<File>, IsComplete>>()
     val needToRestoreFiles get() = galleryApiResult.isInitialized
 
     fun loadLastGallery(driveId: Int, ignoreCloud: Boolean) {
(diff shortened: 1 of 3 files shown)


FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/menu/GalleryViewModel.kt
```diff
@@ -36,28 +36,46 @@ class GalleryViewModel : ViewModel() {
...
-                galleryApiResult.postValue(getLastGallery(driveId, isFirstPage = false, ignoreCloud, currentCursor))
-            }
+            loadLastGallery(driveId, ignoreCloud, isFirstPage = false, currentCursor)
         }
     }
 
-    private fun getLastGallery(
+    fun restoreGalleryFiles() {
+        if (needToRestoreFiles) {
+            galleryApiResult.value = lastGalleryFiles to galleryApiResult.value!!.second
```
tevincent, line 56: Is it something we really want to crash here ? In my opinion, It would be better to display an error with an empty state instead of crashing.

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/ui/menu/GalleryViewModel.kt
```diff
@@ -36,28 +36,46 @@ class GalleryViewModel : ViewModel() {
     private var currentCursor: String? = null
 
     val galleryApiResult = MutableLiveData<Pair<ArrayList<File>, IsComplete>?>()
+    private var lastGalleryFiles = ArrayList<File>()
+
+    val needToRestoreFiles get() = galleryApiResult.isInitialized
```
tevincent, line 41: One thing I like to do is to always group public variables with each other and so on.

app/src/main/java/com/infomaniak/drive/ui/menu/GalleryViewModel.kt
```diff
@@ -36,28 +36,46 @@ class GalleryViewModel : ViewModel() {
     private var currentCursor: String? = null
 
     val galleryApiResult = MutableLiveData<Pair<ArrayList<File>, IsComplete>?>()
+    private var lastGalleryFiles = ArrayList<File>()
```
tevincent, line 39:
```suggestion
    private var lastGalleryFiles = arrayListOf<File>()
```

tevincent (changes requested):

FabianDevel (approved):

tevincent (approved):

PR #1115, opened 2023-11-06
Update Italian translation

mijorus:


JorisBodin:
Thanks
It's also changed on the mail app

PR #1122, opened 2023-11-15
Use `formatShortBinarySize()` to compute sizes

KevinBoulongne:
Depends on #1123

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
+++ b/app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
@@ -28,6 +28,7 @@ import android.graphics.Point
 import android.net.Uri
 import android.os.Build.*
 import android.provider.MediaStore
+import android.text.format.Formatter
 import android.util.DisplayMetrics
 import android.util.Patterns
 import android.view.View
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
```diff
@@ -379,3 +380,40 @@ fun OneTimeWorkRequest.Builder.setExpeditedIfAvailable() = apply {
...
+        valueOnly -> size.binaryToDecimal()
+        else -> size
+    }
+
+    return if (valueOnly) {
+        "$decimalSize"
+    } else {
+        Formatter.formatShortFileSize(this, decimalSize)
+    }
+}
```
tevincent, line 419: It could be great to test this method in order to ensure that we have the right output ?

tevincent (changes requested):
Maybe we can add some tests for the `Context.formatShortBinarySize` method.

app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
KevinBoulongne: Ok, will be done in another PR.

PR #1123, opened 2023-11-15
Clean sizes code

KevinBoulongne:


app/src/main/java/com/infomaniak/drive/views/NotEnoughStorageView.kt
```diff
@@ -75,7 +75,7 @@ class NotEnoughStorageView @JvmOverloads constructor(
         }
     }
 
-    companion object {
-        const val STORAGE_ALERT_MIN_PERCENTAGE = 90
+    private companion object {
+        const val STORAGE_ALERT_MIN_PERCENTAGE = 90.0f
```
tevincent, line 79: Instead of having the whole `companion object` in `private`, you can just put the `const val` as a `private const val.`

tevincent (changes requested):
Private companion object is a bit weird.

tevincent (approved):

PR #1129, opened 2023-11-22
Extract import logic in class

FabianDevel:


app/src/main/java/com/infomaniak/drive/ui/addFiles/AddFileBottomSheetDialog.kt
```diff
@@ -110,12 +109,19 @@ class AddFileBottomSheetDialog : BottomSheetDialogFragment() {
...
+
+        uploadFileHelper = UploadFileHelper(
+            fragment = this@AddFileBottomSheetDialog,
+            parentFolder = currentFolderFile,
+            onOpeningPicker = {
+                trackNewElement("uploadFile")
+                binding.documentUpload.isEnabled = false
+            },
+            onResult = { findNavController().popBackStack() },
+        )
```
tevincent, line 121: Is it possible to initialize uploadFileHelper using a `by lazy` ?

app/src/main/java/com/infomaniak/drive/ui/addFiles/UploadFileHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+        onOpeningPicker: () -> Unit,
+        onResult: (() -> Unit)? = null,
+    ) : this(parentFolder, fragment.findNavController(), onOpeningPicker, onResult) {
+        filePicker = FilePicker(fragment)
+        uploadFilesPermissions = DrivePermissions().apply {
+            registerPermissions(fragment) { authorized -> if (authorized) uploadFiles() }
+        }
+    }
+
+    constructor(
```
tevincent, line 54: It seems that constructor is not used. Maybe it'll be used in another PR ?

app/src/main/java/com/infomaniak/drive/ui/addFiles/UploadFileHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+
+class UploadFileHelper private constructor(
+    private val parentFolder: File,
+    private val navController: NavController,
+    private val onOpeningPicker: () -> Unit,
+    private val onResult: (() -> Unit)?,
+) {
+
+    private lateinit var filePicker: FilePicker
+    private lateinit var uploadFilesPermissions: DrivePermissions
```
tevincent, line 40: `uploadFilesPermissions` or `uploadFilePermissions` ? Sice the class is called `UploadFileHelper`.

app/src/main/java/com/infomaniak/drive/ui/addFiles/UploadFileHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.infomaniak.lib.core.utils.FilePicker
+
+class UploadFileHelper private constructor(
+    private val parentFolder: File,
+    private val navController: NavController,
+    private val onOpeningPicker: () -> Unit,
+    private val onResult: (() -> Unit)?,
+) {
+
+    private lateinit var filePicker: FilePicker
```
tevincent, line 39: You can put this in the constructor to avoid having a lateinit.

tevincent (changes requested):
Questions

app/src/main/java/com/infomaniak/drive/ui/addFiles/UploadFileHelper.kt
FabianDevel: Yes this is used in the next PR, but I really wanted to extract the logic all at once to have a full vision

app/src/main/java/com/infomaniak/drive/ui/addFiles/UploadFileHelper.kt
FabianDevel: I renamed the class to UploadFileHelper

app/src/main/java/com/infomaniak/drive/ui/addFiles/AddFileBottomSheetDialog.kt
FabianDevel: Only if I dissociate the `parentFolder` argument of the constructor in an init function, but then we would need to avoid forgetting calling this init, thus I prefered to keep it in constructor

tevincent (approved):

PR #1132, opened 2023-12-04
Add the possibility to put files into offline mode using the "Select all" button

tevincent:


tevincent:
https://github.com/Infomaniak/android-kDrive/assets/149579879/07b741a4-a7cb-4ed9-83d6-4d750525fda8

tevincent:
Download of files:

https://github.com/Infomaniak/android-kDrive/assets/149579879/935b18b2-7cd4-4503-9fb2-9b53293c6e27

tevincent:
Download of files completed:

https://github.com/Infomaniak/android-kDrive/assets/149579879/91c24ca4-d20a-4d8c-a83d-90f8236be071

tevincent:
Delete selected files from offline:

https://github.com/Infomaniak/android-kDrive/assets/149579879/4a803b7a-3357-446b-b967-3b1ff187ea93

tevincent:
Select all files with different file status (offline and not offline)

https://github.com/Infomaniak/android-kDrive/assets/149579879/04ebe854-4bc0-4c6f-83c0-cded15614fe9

app/src/main/java/com/infomaniak/drive/extensions/PairExtensions.kt
```diff
@@ -0,0 +1,22 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.drive.extensions
+
+fun <A : Any?, B : Any?> Pair<A?, B?>.letAll(callback: (value1: A, value2: B) -> Unit) {
+    if (first != null && second != null) callback.invoke(first!!, second!!)
+}
```
sirambd, line 22: I don't see the point, and it's not easy to understand unless you come and read the code.
And even if you wanted something like that, you should have used an easier-to-understand name like `bothNotNull` with the keyword `infix` to make it an operation.
Here, this function is only used once, so what's the reason? Why not just use a simple `if`?

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    private fun notifyDownloadCancelled() {
+        Intent().apply {
+            action = TAG
+            LocalBroadcastManager.getInstance(applicationContext).sendBroadcast(this)
+        }
+    }
+
+    private fun saveRemoteData(
```
sirambd, line 229: This function is a duplication, using the existing function

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+                override fun update(bytesRead: Long, contentLength: Long, done: Boolean) {
+                    val currentSystemTimeMillis = System.currentTimeMillis()
+                    if (currentSystemTimeMillis - lastUpdateProgressMillis > MAX_INTERVAL_BETWEEN_PROGRESS_UPDATE_MS) {
+                        lastUpdateProgressMillis = currentSystemTimeMillis
+                        val progress = (bytesRead.toFloat() / contentLength.toFloat() * 100F).toInt()
+                        onProgress(progress)
+                    }
+                }
+            })).build()
+    }
```
sirambd, line 271: Avoid duplication: when a function exists, use it whenever possible, otherwise create a utility, but avoid duplicating functions.

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+            /* icon = */ null,
+            /* title = */ applicationContext.getString(R.string.buttonCancel),
+            /* intent = */ cancelPendingIntent
+        )
+        return applicationContext.downloadProgressNotification().apply {
+            setOngoing(true)
+            setContentTitle(applicationContext.getString(R.string.bulkDownloadNotificationTitleNoProgress))
+            addAction(cancelAction)
+        }
+    }
```
sirambd, line 162: This function already exists in `DownloadWorker`, please avoid duplication

app/src/main/java/com/infomaniak/drive/extensions/CoroutineWorkerExtensions.kt
sirambd, line 1: Instead of doing that, create a class to manage an offline, so you can call it in both workers, and you won't have to make duplicates.

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -316,7 +316,9 @@ class MainActivity : BaseActivity() {
 
         launchAllUpload(drivePermissions)
 
-        if (!mainViewModel.ignoreSyncOffline) launchSyncOffline() else mainViewModel.ignoreSyncOffline = false
+        mainViewModel.checkBulkDownloadStatus { isRunning ->
+            if (!isRunning && !mainViewModel.ignoreSyncOffline) launchSyncOffline() else mainViewModel.ignoreSyncOffline = false
+        }
```
sirambd, line 321: Rewrite its own way in MainViewModel, it alone is enough to manage this

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -361,23 +362,47 @@ class MainViewModel(appContext: Application) : AndroidViewModel(appContext) {
...
+            val workQuery = WorkQuery.Builder
+                .fromUniqueWorkNames(arrayListOf(BulkDownloadWorker.TAG))
+                .addStates(arrayListOf(WorkInfo.State.RUNNING))
+                .build()
+            val workInfoList = WorkManager.getInstance(getApplication()).getWorkInfos(workQuery).await()
+            val isRunning = workInfoList.isNotEmpty() && workInfoList.first().state == WorkInfo.State.RUNNING
+            onWorkRunningResult(isRunning)
+        }
+    }
+
```
sirambd, line 383: Move to its own worker and re-write `checkBulkDownloadStatus`

app/src/main/java/com/infomaniak/drive/ui/fileList/FileListFragment.kt
```diff
@@ -556,19 +558,36 @@ open class FileListFragment : MultiSelectFragment(MATOMO_CATEGORY), SwipeRefresh
...
+        val workInfos = workInfoList.filter { it.state == WorkInfo.State.RUNNING }
+
+        if (workInfos.isEmpty()) {
+            updateVisibleProgresses()
+            return
+        }
 
-            val fileId: Int = workInfo.progress.getInt(DownloadWorker.FILE_ID, 0)
-            if (fileId == 0) return@observe
+        workInfos.forEach { workInfo ->
```
sirambd, line 586: Why a foreach ?

sirambd (changes requested):

app/src/main/java/com/infomaniak/drive/extensions/PairExtensions.kt
tevincent: This is for people who want to use that extension to avoid checking the first and second value of a Pair. I think Isaw some of that check but I didn't want to change everywhere we did that check.

I named it `letAll` to make it a bit like `let` to tell `this extension is to check the nullability of both values` of the Pair.

Anyway, if you don't like it. I'll remove it.

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -440,4 +469,11 @@ class MainViewModel(appContext: Application) : AndroidViewModel(appContext) {
         realm.close()
         super.onCleared()
     }
+
+    data class FileRequest(
```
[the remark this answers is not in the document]
tevincent: Because in some cases, it's not a real api call we're doing. `FileResponse` suppose we have a response from a WS. Both of them are fine to me so if you prefer FileResponse, I'll change.

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.Response
+import java.io.BufferedInputStream
+import java.io.File as IOFile
+
+class BulkDownloadWorker(context: Context, workerParams: WorkerParameters) : CoroutineWorker(context, workerParams) {
+
+    private var notificationManagerCompat: NotificationManagerCompat = NotificationManagerCompat.from(applicationContext)
+    private var filesPair: MutableMap<Int, Pair<File?, IOFile?>> = mutableMapOf()
```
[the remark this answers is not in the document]
tevincent: I'm not sure when. I looked at the existing logic we had and it was nullable. Maybe that's not necessary ?

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+        val okHttpClient = AccountUtils.getHttpClient(userDrive.userId, null)
+        val response = downloadFileResponse(
+            fileUrl = ApiRoutes.downloadFile(file),
+            okHttpClient = okHttpClient
+        ) { progress ->
+            if (!isActive) {
+                notificationManagerCompat.cancel(NOTIFICATION_ID)
+                throw CancellationException()
+            }
+            launch(Dispatchers.Main) { setProgress(workDataOf(PROGRESS to progress, FILE_ID to file.id)) }
```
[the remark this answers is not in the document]
tevincent: From my investigation, if we don't do that, not all progress are sent so we cannot update correctly the progress.

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+            SentryLog.d(TAG, "download $progress%")
+        }
+
+        saveRemoteData(response, offlineFile) {
+            launch(Dispatchers.Main) { setProgress(workDataOf(PROGRESS to 100, FILE_ID to file.id)) }
```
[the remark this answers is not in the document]
tevincent: Same as above.

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+    private fun updateDownloadNotification(contentTitle: String, contentText: String, progressPercent: Int) {
+        downloadProgressNotification.apply {
+            setContentTitle(contentTitle)
+            setContentText(contentText)
+            setProgress(100, progressPercent, false)
+            notificationManagerCompat.notifyCompat(applicationContext, NOTIFICATION_ID, build())
+        }
+    }
+
+    private suspend fun startOfflineDownload(file: File, offlineFile: IOFile): Result = withContext(Dispatchers.Default) {
```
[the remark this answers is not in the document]
tevincent: Since we're making a network call, I need to change that to `IO`.

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
tevincent: I agree with you, I'll merge what's existing in `DownloadWorker` and `BulkDownloadWorker`.

app/src/main/java/com/infomaniak/drive/extensions/CoroutineWorkerExtensions.kt
tevincent: Isn't that the same as using an extension ? What's the advantages of using a class ? Since I can create an extension for every method we have in common in `DownloadWorker` and `BulkDownloadWorker` ?

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
```diff
@@ -0,0 +1,286 @@
...
+        }
+    }
+
+    private fun clearFiles() {
+        for ((_, value) in filesPair.entries) {
+            Pair(value.first, value.second).letAll { file, offlineFile ->
+                if (offlineFile.exists() && !file.isIntactFile(offlineFile)) offlineFile.delete()
+            }
+        }
+    }
```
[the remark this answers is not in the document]
tevincent: What do you mean by checking the current ?

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
tevincent: I agree for `downloadFileResponse` but `downloadProgressInterceptor` is different because we do not want to update the progress too often. I'll move `downloadFileResponse` to `CoroutineWorkerExtensions`

app/src/main/java/com/infomaniak/drive/extensions/PairExtensions.kt
sirambd: Yes, remove it, when we need it we'll do it but differently

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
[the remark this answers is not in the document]
sirambd: The current file, in other words, the last file that you tried to download, for those past and future we do not need to check them since you are processing them synchronously

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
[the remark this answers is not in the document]
sirambd: FileRequest sounds too much like a request api, whereas here you need it as a response.
I don't know if we can find better, but in itself `FileResponse` for me it simply means the result of the file, wait maybe we can call it `FileResult`, I find that it works better

app/src/main/java/com/infomaniak/drive/extensions/CoroutineWorkerExtensions.kt
sirambd: The extensions are exposed publicly, but here it is specific for two workers.
And since these two classes have a lot of code in common, then encapsulating everything in a class so that it is reusable in both workers would be quite clean and would remove a lot of duplication.

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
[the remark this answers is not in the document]
sirambd: Ok, but in the logic of your code, it is not possible for it to be nullable.
It's useless because you already manage when it's null you download it from the API otherwise an exception is thrown

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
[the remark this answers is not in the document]
sirambd: It's because you are in a default Coroutine, fix it and normally it should be fine

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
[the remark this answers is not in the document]
sirambd: In an IO dispatcher it should work so we won't need that

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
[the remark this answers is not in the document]
sirambd: It also fixes the progressions

app/src/main/java/com/infomaniak/drive/data/services/BulkDownloadWorker.kt
sirambd: I understand, but you can still use the same method and just ignore the progression when it's not necessary since that's the only difference

app/src/main/java/com/infomaniak/drive/ui/fileList/FileListFragment.kt
tevincent: Yes you're right, I only receive 1 `WorkInfo` every time here.

app/src/main/java/com/infomaniak/drive/utils/Utils.kt
```diff
@@ -286,6 +288,33 @@ object Utils {
         workManager.enqueueUniqueWork(DownloadWorker.TAG, ExistingWorkPolicy.APPEND_OR_REPLACE, downloadRequest)
     }
 
+    fun downloadAsOfflineFiles(context: Context, files: List<File>, userDrive: UserDrive = UserDrive(), onSuccess: () -> Unit) = liveData {
+        val workManager = WorkManager.getInstance(context)
+
+        if (files.any { it.isPendingOffline(context) }) workManager.cancelAllWorkByTag(BulkDownloadWorker::class.java.toString())
+        val inputData = workDataOf(
+            BulkDownloadWorker.FILE_IDS to files.map { it.id }.toIntArray(),
```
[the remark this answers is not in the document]
tevincent: The limitation seems to be 10KB. Even though 10KB of ids seems to be a lot of ids, we can handle the case. I see 2 possibilities:

1. Saving IDs in DB, getting a URI and then passing this one to the Worker using the `setInputData` method
2. Saving that list of IDs somewhere like a ViewModel, a Singleton class etc and then accessing it directly in the Worker

Do you see any other possibilities ?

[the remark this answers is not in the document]
tevincent: One issue with your solution, which is tagging as offline all selected files first and then, getting them from the worker to avoid passing a large list of IDs, is that the file list update itself as downloaded (green mark) when we mark them as offline. So we don't have any progress etc on each file.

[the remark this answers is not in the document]
tevincent: Getting files to put as offline using a new field in File is working great. I have also created a temporary realm migration that will need to be changed later. Tested with 50k files, it's working great. Sometimes, we have a CancellationException because of a time out but the worker is restarted and continue to download the remaining files.

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -149,6 +149,45 @@ object FileController {
...
     }
 
+    fun getFolderOfflineFilesId(folderId: Int): List<Int> {
+        return getRealmInstance().use { realm ->
+            realm.where(File::class.java)
+                .equalTo(File::parentId.name, folderId)
+                .equalTo(File::isMarkedAsOffline.name, true)
+                .getSortQueryByOrder(SortType.NAME_AZ)
+                .findAll()
+                .let { realmFiles ->
```
sirambd, line 159: `let` is useless

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -149,6 +149,45 @@ object FileController {
...
+                        .equalTo(File::id.name, fileId)
+                        .findFirst()?.let { file ->
+                            file.isMarkedAsOffline = isMarkedAsOffline
+                        }
+                }
+            }
+        }
+    }
+
+    fun isFileMarkedAsOffline(customRealm: Realm = getRealmInstance(), fileId: Int, isMarkedAsOffline: Boolean) {
```
sirambd, line 179: Rename to `markFileAsOffline`

app/src/main/java/com/infomaniak/drive/ui/fileList/FavoritesFragment.kt
```diff
@@ -102,7 +103,14 @@ class FavoritesFragment : FileListFragment() {
         // API doesn't support bulk operations for files originating from
         // different parent folders, so we repeat the action for each file.
         // Hence the `areAllFromTheSameFolder` set at false.
-        super.performBulkOperation(type, false, allSelectedFilesCount, destinationFolder, color)
+        super.performBulkOperation(
+            type,
+            folderId,
+            false,
```
sirambd, line 109: Add field name

app/src/main/java/com/infomaniak/drive/ui/fileList/multiSelect/MultiSelectFragment.kt
```diff
@@ -298,20 +315,57 @@ abstract class MultiSelectFragment(private val matomoCategory: String) : Fragmen
...
+        folderId: Int?,
+        mediator: MediatorLiveData<Pair<Int, Int>>,
+    ) {
+        if (folderId != null) {
+            mainViewModel.notificationPermission.checkNotificationPermission(requestPermission = true)
+            mediator.addSource(
+                downloadAsOfflineFiles(
+                    context = requireContext(),
+                    folderId = folderId,
+                    onSuccess = { onIndividualActionSuccess(type, Any()) }
```
sirambd, line 329: Replace `Any` with nullable

app/src/main/java/com/infomaniak/drive/ui/fileList/multiSelect/MultiSelectFragment.kt
```diff
@@ -298,20 +315,57 @@ abstract class MultiSelectFragment(private val matomoCategory: String) : Fragmen
...
+
+    private fun sendRemoveOfflineAction(
+        type: BulkOperationType,
+        selectedFiles: List<File>,
+        mediator: MediatorLiveData<Pair<Int, Int>>,
+    ) {
+        mediator.addSource(
+            mainViewModel.removeSelectedFilesFromOffline(
+                files = selectedFiles,
+                onSuccess = { onIndividualActionSuccess(type, Any()) }
```
sirambd, line 344: Replace `Any` with nullable

app/src/main/java/com/infomaniak/drive/ui/fileList/multiSelect/MultiSelectFragment.kt
```diff
@@ -440,27 +495,12 @@ abstract class MultiSelectFragment(private val matomoCategory: String) : Fragmen
...
-                addSelectedFileToOffline(file, offlineFile, cacheFile)
-            } else {
-                removeSelectedFileFromOffline(file, offlineFile, cacheFile)
-            }
-            onIndividualActionSuccess(type, Unit)
-            closeMultiSelect()
-        }
-    }
-
     private fun addSelectedFileToOffline(file: File, offlineFile: IOFile?, cacheFile: IOFile) {
```
sirambd, line 498: Never used

app/src/main/java/com/infomaniak/drive/ui/menu/FileSubTypeListFragment.kt
```diff
@@ -52,7 +53,14 @@ open class FileSubTypeListFragment : FileListFragment() {
         // API doesn't support bulk operations for files originating from
         // different parent folders, so we repeat the action for each file.
         // Hence the `areAllFromTheSameFolder` set at false.
-        super.performBulkOperation(type, false, allSelectedFilesCount, destinationFolder, color)
+        super.performBulkOperation(
+            type,
+            folderId,
+            false,
```
sirambd, line 59: Add field name

app/src/main/java/com/infomaniak/drive/ui/menu/GalleryFragment.kt
```diff
@@ -264,7 +265,7 @@ class GalleryFragment : MultiSelectFragment(MATOMO_CATEGORY), NoItemsLayoutView.
         // API doesn't support bulk operations for files originating from
         // different parent folders, so we repeat the action for each file.
         // Hence the `areAllFromTheSameFolder` set at false.
-        super.performBulkOperation(type, false, allSelectedFilesCount, destinationFolder, color)
+        super.performBulkOperation(type, folderId, false, allSelectedFilesCount, destinationFolder, color)
```
sirambd, line 268: Add field name

app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
```diff
@@ -0,0 +1,316 @@
...
+class DownloadOfflineFileManager(
+    private val userDrive: UserDrive,
+    private val filesCount: Int,
+    private val downloadWorker: BaseDownloadWorker,
+    private val notificationManagerCompat: NotificationManagerCompat
+) {
+    private var lastDownloadedFile: IOFile? = null
+    private var filesDownloaded = 0
+
+    private var result: ListenableWorker.Result = ListenableWorker.Result.failure()
```
sirambd, line 66: Useless

app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
```diff
@@ -0,0 +1,316 @@
...
+            getFileFromRemote(context, fileId, userDrive) { downloadedFile ->
+                downloadedFile.getOfflineFile(context, userDrive.driveId)?.let { offlineFile ->
+                    lastDownloadedFile = offlineFile
+                }
+            }
+        } else {
+            lastDownloadedFile = offlineFile
+        }
+
+        result = file?.let {
```
sirambd, line 103: Return directly here without `variable`

app/src/main/java/com/infomaniak/drive/utils/NotificationPermission.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.infomaniak.drive.R
+import com.infomaniak.drive.utils.DrivePermissions.Companion.resultPermissions
+import com.infomaniak.lib.core.utils.hasPermissions
+
+class NotificationPermission {
+
+    private lateinit var registerForActivityResult: ActivityResultLauncher<Array<String>>
+    private lateinit var activity: FragmentActivity
+
+    fun registerPermissions(fragment: Fragment, onPermissionResult: ((authorized: Boolean) -> Unit)? = null) {
```
sirambd, line 35: Rename to `registerPermissions`

app/src/main/java/com/infomaniak/drive/utils/NotificationPermission.kt
```diff
@@ -0,0 +1,59 @@
...
+            activity.hasPermissions(notificationPermission) -> true
+            else -> {
+                if (requestPermission) registerForActivityResult.launch(notificationPermission)
+                false
+            }
+        }
+    }
+
+    companion object {
+        val notificationPermission = arrayOf(Manifest.permission.POST_NOTIFICATIONS)
```
sirambd, line 57: Make it private

sirambd (changes requested):

app/src/main/java/com/infomaniak/drive/data/services/BaseDownloadWorker.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    abstract fun isCanceled()
+
+    abstract fun isFinished()
+
+    abstract fun isForOneFile(): Boolean
+
+    abstract fun workerTag(): String
+
+    data class DownloadNotification(
```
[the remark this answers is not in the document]
tevincent: Yes it's still necessary.

sirambd (approved):

PR #1133, opened 2023-12-06
Fix binding npe

FabianDevel:


app/src/main/java/com/infomaniak/drive/ui/addFiles/CreateFolderFragment.kt
```diff
@@ -38,21 +39,34 @@ import com.infomaniak.drive.ui.MainViewModel
...
 
     protected val newFolderViewModel: NewFolderViewModel by navGraphViewModels(R.id.newFolderFragment)
     protected val mainViewModel: MainViewModel by activityViewModels()
+
     protected lateinit var adapter: PermissionsAdapter
 
+    private val folderNameTextWatcher by lazy {
+        object : TextWatcher {
+            override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) = Unit
+            override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) = toggleCreateFolderButton()
```
LunarX, line 63: Just to be sure do we also check if the binding is null inside `toggleCreateFolderbutton()` and send a Sentry if it's the case? And only execute code if binding is set

LunarX (changes requested):

app/src/main/java/com/infomaniak/drive/ui/addFiles/CreateFolderFragment.kt
FabianDevel: This looks like over engineering, this should fix this already exceptionnal error case, and we only have 2 sentries about it. I'd rather not putting potentially useless code that we'll forget removing just in case

LunarX (approved):

PR #1131, opened 2023-11-29
Handle password protected PDF

tevincent:
depends on https://github.com/Infomaniak/android-pdfview/pull/6

PR #1134, opened 2023-12-08
Personalize handle page pdf

tevincent:
Depends on https://github.com/Infomaniak/android-kDrive/pull/1131


tevincent:
https://github.com/Infomaniak/android-kDrive/assets/149579879/284bb8ae-d67a-43db-965d-9f151e8d4e65

PR #1135, opened 2023-12-12
Clean UI settings

FabianDevel:


app/src/main/java/com/infomaniak/drive/data/models/UiSettings.kt
LunarX, line 1: You can copy the inner functions from mail what will prevent you from passing `sharedPreferences` as a receiver each time you call `sharedValue()`

LunarX (changes requested):

app/src/main/java/com/infomaniak/drive/data/models/UiSettings.kt
FabianDevel: I'm not really a fan of these functions, it adds another layer of complexity where this is not really needed
Plus in mail there are a lot of settings, but there's less in kDrive

LunarX: Just to be sure, you'd just copy the ones we need in this file not all of the possible ones.

To me it doesn't add any layer of complexity. You just spare your time by not having to write `sharedPrefences.` every time because the function does it for you like a shortcut. You also don't need to care about that `sharedPrefences` instance anymore once it's setup.

You want a value stored in a shared pref? Give it a key and a default value and you're done. Nothing more, nothing less. Non need to bother with anything else

FabianDevel: To me it's just a sugar where I don't really see its usefulness as with autocompletion everything already writes itself, I find it not worth to create a wrapper to economize something like writing 2 characters and competing it,
then duplicating lines

FabianDevel: After team consultation : this is considered overkill

app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
```diff
@@ -76,6 +77,8 @@ class SaveExternalFilesActivity : BaseActivity() {
         java.io.File(cacheDir, SHARED_FILE_FOLDER).apply { if (!exists()) mkdirs() }
     }
 
+    private val uiSettings by lazy { UiSettings(this) }
```
KevinBoulongne, line 80:
```suggestion
    private val uiSettings by lazy { UiSettings(context = this) }
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

JorisBodin (approved):

PR #1161, opened 2024-01-02
Update to new Tonal Surface color API

FabianDevel:


FabianDevel:
Not necessary because the App theme is still Material2

PR #1225, opened 2024-02-23
Centralize and factorize upload in progress logic

LunarX:


app/src/main/res/values/attrs.xml
```diff
@@ -17,6 +17,10 @@
   -->
 <resources>
 
+    <declare-styleable name="CommonAttributes">
```
tevincent, line 20: Is this really used?

tevincent (changes requested):

app/src/main/res/values/attrs.xml
LunarX: Attributes with the same name can't be declared twice in the same attrs.xml file, so here I need to declare it once and reference it at the two places that need it

tevincent (approved):

PR #1230, opened 2024-02-28
Update API calls to receive `null` avatars if the users have not set any

LunarX:
If we merge this right now, it will work for the user profile's avatar but not the others, maybe we want to wait for the kDrive api to be released even though the current api calls won't fail. 

If so, make it a draft after reviewing it I guess

Edit: All is good now, we can merge.

Depends on https://github.com/Infomaniak/android-core/pull/145

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
@@ -783,11 +783,11 @@ object FileController {
 
     fun getFolderActivities(folder: File, page: Int, userDrive: UserDrive? = null): Map<out Int, FileActivity> {
         return getRealmInstance(userDrive).use { realm ->
-            getFolderActivitiesRec(realm, folder, page, userDrive)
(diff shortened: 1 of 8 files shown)


tevincent (approved):

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -783,11 +783,11 @@ object FileController {
 
     fun getFolderActivities(folder: File, page: Int, userDrive: UserDrive? = null): Map<out Int, FileActivity> {
         return getRealmInstance(userDrive).use { realm ->
-            getFolderActivitiesRec(realm, folder, page, userDrive)
+            getFolderActivitiesRecursive(realm, folder, page, userDrive)
         }
     }
 
-    private fun getFolderActivitiesRec(
+    private fun getFolderActivitiesRecursive(
```
tevincent, line 790: Maybe we can add `tailrec` to this function since it is recursive for memory optimisation?

tevincent (changes requested):

tevincent (approved):

PR #1237, opened 2024-03-04
Modify SwitchAccount button so it doesn't overlap with Search menu in RootFilesFragment

LunarX:
The text below the transparent button needs to still be displayed the exact same way, so we overlay a transparent button to reduce its size independently. By limiting its size we prevent wide switch drive buttons from overlapping with the search menu from the toolbar in RootFilesFragment blocking some clicks that were aimed at the search menu button

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
+++ b/app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
@@ -30,10 +30,7 @@ import android.os.Build.VERSION
 import android.os.Build.VERSION_CODES
 import android.provider.MediaStore
 import android.text.format.Formatter
-import android.transition.AutoTransition
-import android.transition.TransitionManager
-import android.transition.TransitionSet
-import android.util.DisplayMetrics
+import android.transition.*
 import android.util.Patterns
 import android.view.ViewGroup
 import android.view.animation.Animation
@@ -46,7 +43,7 @@ import androidx.core.view.children
 import androidx.core.view.isGone
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
-import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.Lifecycle.*
 import androidx.lifecycle.LifecycleEventObserver
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.LiveData
@@ -405,7 +402,8 @@ fun LayoutSwitchDriveBinding.setDriveHeader(currentDrive: Drive) {
 }
 
 fun LayoutSwitchDriveBinding.setupSwitchDriveButton(fragment: Fragment) {
-    AccountUtils.getCurrentDrive()?.let { setDriveHeader(it) }
+
+    AccountUtils.getCurrentDrive()?.let(::setDriveHeader)
 
     if (DriveInfosController.hasSingleDrive(AccountUtils.currentUserId)) {
         switchDriveButton.apply {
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/layout_switch_drive.xml
```diff
@@ -45,6 +45,49 @@
...
         app:iconTint="@color/iconColor"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintHorizontal_bias="0"
+        app:layout_constraintStart_toStartOf="parent"
         tools:text="Drive de test dev infomaniak" />
 
-</LinearLayout>
+    <!--  To prevent the button from overlapping the toolbar search button in the RootFilesFragment we do all of this complex
+    computations with views -->
```
KevinBoulongne, line 55: There is a `marginEnd` or something similar missing somewhere.

KevinBoulongne (changes requested):

KevinBoulongne (changes requested):

app/src/main/res/layout/layout_switch_drive.xml
LunarX: The fact that the ripple stops before the end of the button in extrem cases is actually the intended behavior of this PR

app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
```diff
@@ -430,21 +429,22 @@ fun LayoutSwitchDriveBinding.setDriveHeader(currentDrive: Drive) {
...
+
+    if (DriveInfosController.hasSingleDrive(AccountUtils.currentUserId)) {
+        switchDriveButton.apply {
             icon = null
             isEnabled = false
-        } else {
-            setOnClickListener { fragment.safeNavigate(R.id.switchDriveDialog) }
         }
+    } else {
+        offsetOverlayedRipple.setOnClickListener { fragment.safeNavigate(R.id.switchDriveDialog) }
```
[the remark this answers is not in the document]
LunarX: The issue should not be cause by this PR and should be present beforehand. The issue seems to be caused by the animation of the app when we restart it entirely which looks kinda hard to fix

app/src/main/res/layout/layout_switch_drive.xml
```diff
@@ -15,12 +15,13 @@
...
   ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   -->
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="120.8dp"
-    android:gravity="bottom"
+    android:paddingStart="16dp"
```
tevincent, line 23: You can use `marginStandardMedium` for that no?

tevincent (changes requested):

app/src/main/res/layout/layout_switch_drive.xml
LunarX: This value here should truly never change no matter what our `marginStandardMedium` actual value becomes, so it's hard to say

KevinBoulongne (approved):

tevincent (approved):

PR #1255, opened 2024-03-20
Feat: Drive init v2

sirambd:


app/src/main/java/com/infomaniak/drive/utils/AccountUtils.kt
```diff
@@ -157,16 +157,17 @@ object AccountUtils : CredentialManager() {
...
     }
 
-    private fun reloadAppIfNeeded(fromMaintenance: Boolean, driveInfo: DriveInfo) {
+    private suspend fun reloadAppIfNeeded(fromMaintenance: Boolean, driveInfo: DriveInfo) {
+        val internalDrives = driveInfo.drives.filter { it.isDriveUser() }
         if (fromMaintenance) {
-            if (driveInfo.drives.main.any { drive -> !drive.maintenance }) {
-                GlobalScope.launch(Dispatchers.Main) {
+            if (internalDrives.any { drive -> !drive.maintenance }) {
+                Dispatchers.Main {
```
tevincent, line 164: We could put that in another method since we have the same lines line 170 - 172

tevincent (changes requested):

tevincent (approved):

PR #1260, opened 2024-03-27
Clean OnItemClickListener by adding super call and add Unit to force implementation

tevincent:


LunarX (commented):
Wouldn't it make sens to move the interface out of this class and into its own file as well?

Also wouldn't there be a benefit to already factorizing a bit the actions you plan on using in your next PR so they're easier to implement then?

PR #1262, opened 2024-03-31
Fix fdroid image

yoriiis:
Fix source image

image

The image source is get from https://github.com/Infomaniak/android-kMail/blob/master/README.md?plain=1#L6C12-L6C64


JorisBodin (approved):

JorisBodin:
Thank you

PR #1264, opened 2024-04-02
Fix: (ApiV3) - After the root you lose folder data

sirambd:
When you return to the root of the files, the data is downloaded again even though the folders are complete.
Depends on #1263

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
@@ -640,12 +640,13 @@ object FileController {
     }
 
     private fun keepSubFolderChildren(localFolderChildren: List<File>?, remoteFolderChildren: List<File>) {
-        val oldChildren = localFolderChildren?.filter { it.children.isNotEmpty() || it.isOffline }?.associateBy { it.id }
+        val oldChildren = localFolderChildren?.filter { it.isFolder() || it.isOffline }?.associateBy { it.id }
         remoteFolderChildren.forEach { newFile ->
             oldChildren?.get(newFile.id)?.let { oldFile ->
                 newFile.apply {
-                    if (oldFile.isFolder()) children = oldFile.children
+                    if (oldFile.isFolder()) keepOldLocalFilesData(oldFile, newFile)
                     isOffline = oldFile.isOffline
+
                 }
             }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -640,12 +640,13 @@ object FileController {
...
     private fun keepSubFolderChildren(localFolderChildren: List<File>?, remoteFolderChildren: List<File>) {
-        val oldChildren = localFolderChildren?.filter { it.children.isNotEmpty() || it.isOffline }?.associateBy { it.id }
+        val oldChildren = localFolderChildren?.filter { it.isFolder() || it.isOffline }?.associateBy { it.id }
         remoteFolderChildren.forEach { newFile ->
             oldChildren?.get(newFile.id)?.let { oldFile ->
                 newFile.apply {
-                    if (oldFile.isFolder()) children = oldFile.children
+                    if (oldFile.isFolder()) keepOldLocalFilesData(oldFile, newFile)
                     isOffline = oldFile.isOffline
+
```
tevincent, line 649: You have an empty line here.

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -753,6 +754,7 @@ object FileController {
     private fun keepOldLocalFilesData(oldFile: File, newFile: File) {
         newFile.apply {
             children = oldFile.children
+            cursor = oldFile.cursor
```
tevincent, line 757: We could improve that function like this maybe?

```kotlin
private fun keepOldLocalFilesData(oldFile: File, newFile: File) = with(newFile) {
        children = oldFile.children
        cursor = oldFile.cursor
        isComplete = oldFile.isComplete
        isOffline = oldFile.isOffline
        responseAt = oldFile.responseAt
        versionCode = oldFile.versionCode
    }
```

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
sirambd: I agree with the improvement, but in this kind of case it's more of an extension, avoiding the dupplication of `newFile`.

tevincent (approved):

PR #1280, opened 2024-04-16
fix: Root files blink

sirambd:


LunarX (commented):
This is definitely way better than what was previously written and should have been like this from the get go. But this won't help with files blinking when first opening the MainActivity, so ... is this enough?

LunarX (approved):

PR #1293, opened 2024-04-30
Fix visibility of dialog when we upload files

tevincent:


app/src/main/java/com/infomaniak/drive/ui/addFiles/ImportFilesDialog.kt
```diff
@@ -109,7 +108,7 @@ class ImportFilesDialog : DialogFragment() {
         }
         lifecycleScope.launch {
```
sirambd, line 109: No need, you can delete this `launch`

sirambd (changes requested):

sirambd (approved):

PR #1302, opened 2024-05-07
fix: (Search) ApiV3 - Search bar not displayed for the first time

sirambd:


app/src/main/java/com/infomaniak/drive/ui/fileList/SearchFragment.kt
```diff
@@ -212,7 +212,7 @@ class SearchFragment : FileListFragment() {
     }
 
     private fun configureRecentSearches() = with(binding) {
-        fileListLayout.addView(recentSearchesBinding.root, 1)
+        fileListLayout.addView(recentSearchesBinding.root, 2)
```
LunarX, line 215: So that's the fix. This is so error prone actually. Shouldn't we include the layout directly inside fragment_file_list.xml and only show it when needed instead?

LunarX (approved):

tevincent (changes requested):

tevincent (approved):

PR #1307, opened 2024-05-14
fix: Create a dropbox from a sharedWithMe logout user

sirambd:
As on `iOS`, you can't create a dropbox from sharedWithMe because it doesn't work, and it disconnects the user.

Depends on #1306

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/addFiles/NewFolderFragment.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/addFiles/NewFolderFragment.kt
@@ -82,13 +82,15 @@ class NewFolderFragment : Fragment() {
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/ui/addFiles/NewFolderFragment.kt
```diff
@@ -81,8 +82,9 @@ class NewFolderFragment : Fragment() {
         }
     }
 
-    private fun initDropBoxFolder(drive: Drive?) {
-        binding.dropBox.setOnClickListener {
+    private fun initDropBoxFolder(drive: Drive?) = binding.dropBox.apply {
```
FabianDevel, line 85: I like to put a `:unit` when an apply return something, and the method should not return anything

FabianDevel (changes requested):

FabianDevel (approved):

PR #1312, opened 2024-05-23
Fix wrong snackbar being displayed when importing files on low memory

TommyDL-Infomaniak:
Before:

https://github.com/Infomaniak/android-kDrive/assets/170427967/5cbdf4df-190a-4788-af23-54debb7e8b1b

After:

https://github.com/Infomaniak/android-kDrive/assets/170427967/c4b78c11-8a62-4c3a-88c8-0044fe83f716

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/addFiles/ImportFilesDialog.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/addFiles/ImportFilesDialog.kt
@@ -74,12 +74,13 @@ class ImportFilesDialog : DialogFragment() {
         val errorCount = importCount - successCount
 
         if (errorCount > 0) {
-            if (isLowMemory()) {
-                showSnackbar(R.string.uploadOutOfMemoryError, showAboveFab = true)
+            val errorMessage = if (isLowMemory()) {
+                getString(R.string.uploadOutOfMemoryError)
             } else {
-                val errorMessage = resources.getQuantityString(R.plurals.snackBarUploadError, errorCount, errorCount)
-                showSnackbar(errorMessage, showAboveFab = true)
+                resources.getQuantityString(R.plurals.snackBarUploadError, errorCount, errorCount)
             }
+
+            showSnackbar(errorMessage, showAboveFab = true)
         }
 
         if (successCount > 0) mainViewModel.refreshActivities.value = true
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/ui/addFiles/ImportFilesDialog.kt
```diff
@@ -73,7 +73,9 @@ class ImportFilesDialog : DialogFragment() {
         super.onDismiss(dialog)
         val errorCount = importCount - successCount
 
-        if (errorCount > 0) {
+        if (isLowMemory()) {
+            showSnackbar(R.string.uploadOutOfMemoryError, showAboveFab = true)
+        } else if (errorCount > 0) {
```
LunarX, line 78: If there's no importation error, the user will still get a out of memory snackbar regardless, which doesn't make much sens

LunarX (changes requested):

app/src/main/java/com/infomaniak/drive/ui/addFiles/ImportFilesDialog.kt
```diff
@@ -117,27 +119,22 @@ class ImportFilesDialog : DialogFragment() {
...
-                            fileName = fileName,
-                            fileSize = outputFile.length(),
-                            remoteFolder = navArgs.folderId,
-                            type = UploadFile.Type.UPLOAD.name,
-                            userId = AccountUtils.currentUserId,
-                        ).store()
-                        successCount++
-                        currentImportFile = null
-                    }
+                if (!isLowMemory()) {
```
LunarX, line 122: We could leave early using the following code instead of having to write the `if` as is

``` kt
if (isLowMemory()) return@withContext
```

And we can also avoid computing the file name and the files dates if are not going to use them anyway because memory is low

app/src/main/java/com/infomaniak/drive/ui/addFiles/ImportFilesDialog.kt
```diff
@@ -73,7 +73,9 @@ class ImportFilesDialog : DialogFragment() {
         super.onDismiss(dialog)
         val errorCount = importCount - successCount
 
-        if (errorCount > 0) {
+        if (errorCount > 0 && isLowMemory()) {
+            showSnackbar(R.string.uploadOutOfMemoryError, showAboveFab = true)
+        } else if (errorCount > 0) {
```
LunarX, line 78: Maybe we could write a `if` statement that handles checking if `errorCount > 0` and inside we can check with a second `if` statement which snackbar needs to be displayed

FabianDevel (changes requested):
There is a snackbar that appears to be useless

FabianDevel (approved):

LunarX (approved):

PR #1316, opened 2024-05-27
Fix tab inserting the toString of objects instead of autocompletion when sharing folder

TommyDL-Infomaniak:
The problem is mostly solved, pressing tab will not insert the object toString anymore, but it will insert the current text, thus not changing anything. 
A consequence of this is that the UI is refreshed and the list of users that was displayed gets closed.

This is the best solution found that does not require a refactor of the MaterialAutoCompleteTextView.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
+++ b/app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
@@ -217,17 +217,18 @@ fun MaterialAutoCompleteTextView.setupAvailableShareableItems(
     itemList: List<Shareable>,
     notShareableIds: ArrayList<Int> = arrayListOf(),
     notShareableEmails: ArrayList<String> = arrayListOf(),
-    onDataPassed: (item: Shareable) -> Unit
+    onDataPassed: (item: Shareable) -> Unit,
 ): AvailableShareableItemsAdapter {
     setDropDownBackgroundResource(R.drawable.background_popup)
     val availableUsersAdapter = AvailableShareableItemsAdapter(
         context = context,
         itemList = ArrayList(itemList),
         notShareableIds = notShareableIds,
-        notShareableEmails = notShareableEmails
-    ) { item ->
-        onDataPassed(item)
-    }
+        getCurrentText = { text },
+        notShareableEmails = notShareableEmails,
+        onItemClick = onDataPassed,
+    )
+
     setAdapter(availableUsersAdapter)
     handleActionDone { if (text.isNotBlank()) !availableUsersAdapter.addFirstAvailableItem() }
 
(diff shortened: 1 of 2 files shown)


KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
```diff
@@ -224,10 +224,12 @@ fun MaterialAutoCompleteTextView.setupAvailableShareableItems(
         context = context,
         itemList = ArrayList(itemList),
         notShareableIds = notShareableIds,
+        getCurrentText = { return@AvailableShareableItemsAdapter text.toString() },
         notShareableEmails = notShareableEmails
```
KevinBoulongne, line 228:
```suggestion
        notShareableEmails = notShareableEmails,
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
```diff
@@ -224,10 +224,12 @@ fun MaterialAutoCompleteTextView.setupAvailableShareableItems(
         context = context,
         itemList = ArrayList(itemList),
         notShareableIds = notShareableIds,
+        getCurrentText = { return@AvailableShareableItemsAdapter text.toString() },
```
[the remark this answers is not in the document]
FabianDevel: You can even remove do this to simplify : 
```suggestion
        getCurrentText = { text },
```

FabianDevel (changes requested):
This creates a bug when you select a user then press enter

app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
```diff
@@ -217,17 +217,18 @@ fun MaterialAutoCompleteTextView.setupAvailableShareableItems(
...
     val availableUsersAdapter = AvailableShareableItemsAdapter(
         context = context,
         itemList = ArrayList(itemList),
         notShareableIds = notShareableIds,
-        notShareableEmails = notShareableEmails
-    ) { item ->
-        onDataPassed(item)
-    }
+        getCurrentText = { text },
+        notShareableEmails = notShareableEmails,
```
LunarX, line 228: You could reorder these lines to keep the same ordering as inside the function declaration to make things easier to follow

LunarX (changes requested):

TommyDL-Infomaniak:
> This creates a bug when you select a user then press enter

This has the same behavior as using tab.

KevinBoulongne (approved):

LunarX (approved):

PR #1320, opened 2024-05-30
Fix: When the token has expired, pending requests continue to renew the token

sirambd:


app/src/main/java/com/infomaniak/drive/MainApplication.kt
```diff
@@ -185,5 +185,7 @@ class MainApplication : Application(), ImageLoaderFactory, DefaultLifecycleObser
         }
 
         override suspend fun getApiToken(): ApiToken? = AccountUtils.currentUser?.apiToken
+
+        override fun getCurrentUserId() = AccountUtils.currentUserId
```
LunarX, line 189: Do we want to specify the return type like in mail?

LunarX (approved):

PR #1330, opened 2024-06-07
Remove the gap between the bottom sheet and the bottom of the screen

NicolasBourdin88:
The gap is the result of set 'isFitContent' to false on the bottom sheet behavior but unfortunately it's either the gap or the half state on the bottom sheet.

The only solution that i founded is to fill the gap with 'match_parent' so we have half_state and no gap.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/utils/PreviewUtils.kt
+++ b/app/src/main/java/com/infomaniak/drive/utils/PreviewUtils.kt
@@ -49,8 +49,6 @@ fun Activity.setupBottomSheetFileBehavior(
     bottomSheetBehavior.apply {
         isHideable = true
         this.isDraggable = isDraggable
-        // TODO This need to be true for ExternalFileInfoActionsView, otherwise, the BottomSheet can go to the top of the screen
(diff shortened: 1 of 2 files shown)


tevincent:
I want to test something. I'll let you know if I find something.

tevincent:
What we had before:

https://github.com/Infomaniak/android-kDrive/assets/149579879/b691d652-f81a-44de-8d33-f1a7fba049ff

The quickfix:

https://github.com/Infomaniak/android-kDrive/assets/149579879/95fc8ac8-912d-4c47-8a0c-cf692b523b64

With a custom class to handle the BottomSheet bahavior:
https://github.com/Infomaniak/android-kDrive/assets/149579879/9957da8b-e816-418e-bc31-4390694c4a71

LunarX (changes requested):
This removes the native interaction behaviors that the bottom sheet naturally has makes it harder to use and less intuitive unfortunately

KevinBoulongne:
> This removes the native interaction behaviors that the bottom sheet naturally has makes it harder to use and less intuitive unfortunately

What does it remove ? Your comment is very unclear.

PR #1332, opened 2024-06-11
Fix sync bottomsheet which was collapsed on tablet in landscape

tevincent:
Before fix:

https://github.com/Infomaniak/android-kDrive/assets/149579879/b0cecbaa-c197-437b-ae0c-56c52c0c70cf

After fix:

https://github.com/Infomaniak/android-kDrive/assets/149579879/f9a939ec-9bb4-4ef3-892e-0df0920c061c


KevinBoulongne (changes requested):

LunarX (changes requested):
Why do we introduce a difference in behavior between bottom sheets?

This also makes landscape bottom sheets corners square shaped compared to portrait bottom sheets corners rounded shape which is also another type of incoherence.

By the way, in the title you mention tablet mode but this is not only constrained to tablets, it affects phones the same way as tablets.

Finally, a view tree observer global layout listener is quite heavy, is there no simpler and lighter solution?

PR #1358, opened 2024-07-16
Fix several SonarCloud issues

TommyDL-Infomaniak:


app/src/main/java/com/infomaniak/drive/ui/fileList/ConvertToDropBoxFragment.kt
```diff
@@ -44,36 +44,37 @@ class ConvertToDropBoxFragment : ManageDropboxFragment() {
...
+                saveButton.showProgressCatching()
+                mainViewModel.createDropBoxFolder(
+                    file = file,
+                    emailWhenFinished = emailWhenFinishedSwitch.isChecked,
+                    password = if (passwordSwitch.isChecked) passwordTextInput.text?.toString() else null,
+                    limitFileSize = limitFileSize?.let { Utils.convertGigaByteToBytes(it) },
+                    validUntil = if (expirationDateSwitch.isChecked) expirationDateInput.getCurrentTimestampValue() else null
+                ).observe(viewLifecycleOwner) { apiResponse ->
+                    if (apiResponse.isSuccess()) {
+                        mainViewModel.createDropBoxSuccess.value = apiResponse.data
```
tevincent, line 72: We could have just emit a boolean (success or not of the ApiCall) and in the viewModel, we can set the `createDropBoxSuccess` to `apiResponse.data` in case of a success.

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/ui/fileList/ConvertToDropBoxFragment.kt
```diff
@@ -44,36 +44,37 @@ class ConvertToDropBoxFragment : ManageDropboxFragment() {
         shareLinkCardView.isGone = true
         disableButton.isGone = true
 
-        FileController.getFileById(navigationArgs.fileId)?.let { file ->
-            updateUi(file, null)
+        val file = FileController.getFileById(navigationArgs.fileId) ?: return
 
-            settings.expirationDateInput.init(fragmentManager = parentFragmentManager)
+        updateUi(file, dropBox = null)
```
tevincent, line 49: You can just set the default value of `dropBox` to `null` in the `updateUi` method to avoid setting it here.

tevincent (changes requested):

tevincent (approved):

PR #1367, opened 2024-07-25
Migrate from the old exoPlayer to the new Media3 and add the possibility to play media in the background

tevincent:
https://github.com/user-attachments/assets/5eef405d-000b-4b5a-93ad-80a0737cd149


KevinBoulongne (approved):

tevincent:
We also need to clean the code, add a service and add a notification with play/pause.

sirambd (changes requested):
- [ ]The video doesn't stop when you swipe on other media.
- [ ] Video competition problem, when 2 videos are consecutive.

PR #1370, opened 2024-07-31
Fix: Do not display colored folder in preview

FabianDevel:
You can click on the button to color folders while opening file options.
In the preview, nothing happens, but in the FileDetailBottomSheet it opens the color change bottomsheet.

This is a regression


LunarX (approved):

LunarX:
This PR is missing, some other views that also lost their default `gone` visibility. The PR #1373 takes this commit and adds the other missing visibility `gone` as well

PR #1376, opened 2024-08-07
Stop capturing mqtt connection lost errors to sentry

LunarX:
This line was a simple Android log until we introduced SentryLog where it became an exception that got captured. But this case looks very common, no need to spam Sentry with this. This PR stops capturing it into Sentry and only adds it as a breadcrumb like the other mqtt logs.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/services/MqttClientWrapper.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/services/MqttClientWrapper.kt
@@ -145,7 +145,7 @@ object MqttClientWrapper : MqttCallback, LiveData<MqttNotification>() {
     }
 
     override fun connectionLost(cause: Throwable?) {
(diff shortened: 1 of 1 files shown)


TommyDL-Infomaniak (approved):

app/src/main/java/com/infomaniak/drive/data/services/MqttClientWrapper.kt
```diff
@@ -145,7 +145,7 @@ object MqttClientWrapper : MqttCallback, LiveData<MqttNotification>() {
     }
 
     override fun connectionLost(cause: Throwable?) {
-        SentryLog.e("MQTT Error", "Connection has been lost. Stacktrace below.")
+        SentryLog.i("MQTT Error", "Connection has been lost. Stacktrace below.")
```
sirambd, line 148: Adds the cause directly to the sentry log

sirambd (changes requested):

sirambd (approved):

PR #1378, opened 2024-08-08
Can now change accounts by double-tap and hold on the profile icon

TommyDL-Infomaniak:
[untitled.webm](https://github.com/user-attachments/assets/c2f82250-fe9c-474c-8e68-3c55f28fa05f)

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
@@ -32,6 +32,8 @@ import android.os.Bundle
 import android.os.FileObserver
 import android.provider.DocumentsContract
 import android.provider.MediaStore
+import android.view.GestureDetector
+import android.view.MotionEvent
 import android.view.View
 import androidx.activity.result.IntentSenderRequest
 import androidx.activity.result.contract.ActivityResultContracts
@@ -60,7 +62,6 @@ import com.google.android.material.snackbar.Snackbar
 import com.infomaniak.drive.BuildConfig
 import com.infomaniak.drive.GeniusScanUtils.scanResultProcessing
 import com.infomaniak.drive.GeniusScanUtils.startScanFlow
-import com.infomaniak.drive.MatomoDrive.trackAccountEvent
 import com.infomaniak.drive.MatomoDrive.trackEvent
 import com.infomaniak.drive.MatomoDrive.trackInAppReview
 import com.infomaniak.drive.MatomoDrive.trackInAppUpdate
@@ -97,7 +98,6 @@ import com.infomaniak.lib.stores.updatemanagers.InAppUpdateManager
 import io.sentry.Breadcrumb
 import io.sentry.Sentry
 import io.sentry.SentryLevel
-import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -206,6 +208,26 @@ class MainActivity : BaseActivity() {
...
+            if (it.tag != null && Date().time - (it.tag as Date).time < DOUBLE_TAP_DELAY) {
+                CoroutineScope(Dispatchers.IO).launch {
+                    trackAccountEvent("switch")
+                    AccountUtils.switchToNextUser()
+                }.also { navController.navigate(R.id.homeFragment) }
+
+                it.tag = null
+            } else {
+                it.tag = Date()
+            }
```
LunarX, line 228: Instead of bothering with manually computing the time to detect double tap we can use the listeners that are made for this very purpose:

```kt
val gestureDetector = GestureDetector(this@MainActivity, object : SimpleOnGestureListener() {
    override fun onDoubleTap(e: MotionEvent): Boolean {
        // TODO: Switch user
        return true
    }
})

profileItem.setOnTouchListener { v, event -> gestureDetector.onTouchEvent(event) }
```

LunarX, line 228: We should probably also not switch the user on double tap if there's only a single user

LunarX (changes requested):

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
TommyDL-Infomaniak: I came across this when searching for a solution, but could not make it to work. In fact, I was missing the last line. It is now implemented, thank you!

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -202,7 +204,25 @@ class MainActivity : BaseActivity() {
         }
     }
 
+    // We use this SuppressLint because we don't want to execute performClick on profileItem when double tapping.
+    @SuppressLint("ClickableViewAccessibility")
     private fun setupBottomNavigation() = with(binding) {
+
+        val profileItem = bottomNavigation.findViewById<View>(R.id.menuFragment)
+
```
KevinBoulongne, line 212:
```suggestion
```

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -202,7 +204,25 @@ class MainActivity : BaseActivity() {
...
+                mainViewModel.switchToNextUser { navController.navigate(R.id.homeFragment) }
+                return true
+            }
+
+            override fun onLongPress(e: MotionEvent) {
+                navController.navigate(R.id.switchUserActivity)
+            }
+        })
+
+        profileItem.setOnTouchListener { _, event -> gestureDetector.onTouchEvent(event) }
```
KevinBoulongne, line 224:
```suggestion
        bottomNavigation.findViewById<View>(R.id.menuFragment).setOnTouchListener { _, event ->
            gestureDetector.onTouchEvent(event)
        }
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -632,6 +633,17 @@ class MainViewModel(
         }
     }
 
+    fun switchToNextUser(callback: () -> Unit) {
+        viewModelScope.launch(Dispatchers.IO) {
+            if (AccountUtils.getAllUsersSync().size < 2) return@launch
+
+            getContext().trackAccountEvent("switch")
```
FabianDevel, line 640: Maybe we want to track specifically if the user double tap

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -632,6 +633,17 @@ class MainViewModel(
         }
     }
 
+    fun switchToNextUser(callback: () -> Unit) {
```
FabianDevel, line 636: rename the callback to a more meaningful name like "onUserSwitched"

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -632,6 +633,17 @@ class MainViewModel(
...
+        viewModelScope.launch(Dispatchers.IO) {
```
FabianDevel, line 637: Couldnt this feat on 1 line ? 
We almost always write this onelined to gain an indent level

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -598,5 +618,6 @@ class MainActivity : BaseActivity() {
 
     companion object {
         private const val SYNCED_FILES_DELETION_FILES_AMOUNT = 10
+        private const val DOUBLE_TAP_DELAY = 300L
```
FabianDevel, line 621: This seems to be unused

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -202,7 +204,25 @@ class MainActivity : BaseActivity() {
...
+                mainViewModel.switchToNextUser { navController.navigate(R.id.homeFragment) }
+                return true
+            }
+
+            override fun onLongPress(e: MotionEvent) {
+                navController.navigate(R.id.switchUserActivity)
+            }
+        })
+
+        bottomNavigation.findViewById<View>(R.id.menuFragment).setOnTouchListener { _, event ->
```
FabianDevel, line 222: coulnd't this be find by binding ?

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
TommyDL-Infomaniak: I could not access a specific view of the BottomNavigation and set an onTouchListener, but maybe I'm missing something.

FabianDevel (approved):

KevinBoulongne (approved):

PR #1385, opened 2024-08-12
Hide the print option for non PDF file

tevincent:
https://github.com/user-attachments/assets/3603af17-7ff5-4588-9ae2-765a24a55659


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/drive/ui/bottomSheetDialogs/FileInfoActionsBottomSheetDialog.kt
```diff
@@ -95,6 +96,8 @@ class FileInfoActionsBottomSheetDialog : BottomSheetDialogFragment(), FileInfoAc
                 isSharedWithMe = navigationArgs.userDrive.sharedWithMe,
             )
             updateCurrentFile(currentFile)
+
+            binding.fileInfoActionsView.setPrintVisibility(isGone = currentFile.extensionType != ExtensionType.PDF.value)
```
FabianDevel, line 100: Maybe we could add an extension `isPdf` in File.kt to avoid duplicating the logic here. I'm not totally sure it's worth it though.

But we could use the `getFileType()` function to directly check on ExtensionType.PDF without the value.

FabianDevel (changes requested):

FabianDevel (approved):

PR #1391, opened 2024-08-15
Removing Sentry for RefreshTokenError

tevincent:
This can happens when the user did not connect to the app for a long time. We don't have a RefreshToken anymore so this log is not useful anymore.


LunarX:
Did we also remove them on mail? Also what is the reason that makes them not needed anymore?

tevincent:
We don't have that Sentry log in Mail.

PR #1382, opened 2024-08-12
02 Add preview actions to Public shared files

FabianDevel:
Depends on #1381 

This PR add the actions on files in the preview slider fragment

Actions that are added : 
 - open with
 - send a copy
 - save to kDrive (if you are connected)
 - download
 - print (if it can be printed)

PR #1392, opened 2024-08-16
03 Add File shared list actions

FabianDevel:
Depends on #1382 

This PR add the `Download all` button  and the main button to import to your kDrive the files

Important : For now the back isn't ready for the import of the files, so we might be forced to remove this feature in a latter PR if no changes are planned on the backend side.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
@@ -48,6 +48,7 @@ import com.infomaniak.drive.data.models.UploadFile
 import com.infomaniak.drive.data.models.UserDrive
 import com.infomaniak.drive.data.models.drive.Drive
 import com.infomaniak.drive.databinding.ActivitySaveExternalFileBinding
+import com.infomaniak.drive.ui.fileList.FileListFragment
 import com.infomaniak.drive.ui.fileList.SelectFolderActivity
 import com.infomaniak.drive.ui.fileList.SelectFolderActivityArgs
 import com.infomaniak.drive.ui.menu.settings.SelectDriveDialog
@@ -269,6 +270,16 @@ class SaveExternalFilesActivity : BaseActivity() {
         binding.saveButton.apply {
             initProgress(this@SaveExternalFilesActivity)
             setOnClickListener {
+                if (navigationArgs.isPublicShare) {
+                    Intent().apply {
+                        putExtra(DESTINATION_DRIVE_ID_KEY, selectDriveViewModel.selectedDrive.value?.id)
+                        putExtra(DESTINATION_FOLDER_ID_KEY, saveExternalFilesViewModel.folderId.value)
+                        setResult(RESULT_OK, this)
+                    }
+                    finish()
+                    return@setOnClickListener
+                }
+
                 showProgressCatching()
                 if (drivePermissions.checkSyncPermissions()) {
                     val userId = selectedUserId.value!!
@@ -382,7 +393,7 @@ class SaveExternalFilesActivity : BaseActivity() {
     }
 
     private fun isValidFields(): Boolean {
-        return (isMultiple || !binding.fileNameEdit.showOrHideEmptyError()) &&
+        return (isMultiple || !binding.fileNameEdit.showOrHideEmptyError() || navigationArgs.isPublicShare) &&
                 selectDriveViewModel.selectedUserId.value != null &&
                 selectDriveViewModel.selectedDrive.value != null &&
                 saveExternalFilesViewModel.folderId.value != null &&
@@ -520,5 +531,7 @@ class SaveExternalFilesActivity : BaseActivity() {
     companion object {
         const val SHARED_FILE_FOLDER = "shared_files"
         const val LAST_MODIFIED_URI_KEY = "last_modified"
+        const val DESTINATION_DRIVE_ID_KEY  = "destination_drive_id"
+        const val DESTINATION_FOLDER_ID_KEY  = "destination_folder_id"
     }
 }
--- a/app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
@@ -48,7 +48,6 @@ import com.infomaniak.drive.data.models.UploadFile
 import com.infomaniak.drive.data.models.UserDrive
 import com.infomaniak.drive.data.models.drive.Drive
 import com.infomaniak.drive.databinding.ActivitySaveExternalFileBinding
-import com.infomaniak.drive.ui.fileList.FileListFragment
 import com.infomaniak.drive.ui.fileList.SelectFolderActivity
 import com.infomaniak.drive.ui.fileList.SelectFolderActivityArgs
 import com.infomaniak.drive.ui.menu.settings.SelectDriveDialog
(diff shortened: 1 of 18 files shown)


tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedViewModel.kt
```diff
@@ -95,6 +95,24 @@ class FileSharedViewModel(val savedStateHandle: SavedStateHandle) : ViewModel()
         getSharedFilesJob.cancelChildren()
     }
 
+    fun importFilesToDrive(
+        destinationDriveId: Int,
+        destinationFolderId: Int,
+        fileIds: List<Int> = emptyList(),
```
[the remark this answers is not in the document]
FabianDevel: For now yes, as the multiselection is not working yet. So the fileIds and exceptedFileIds will be passed when it will be done

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedListFragment.kt
```diff
@@ -41,8 +55,29 @@ class FileSharedListFragment : FileListFragment() {
...
     override var hideBackButtonWhenRoot: Boolean = false
 
+    private var drivePermissions: DrivePermissions? = null
+    private val selectDriveAndFolderResultLauncher = registerForActivityResult(StartActivityForResult()) {
+        it.whenResultIsOk { data ->
+            if (data == null) {
+                showSnackbar(RCore.string.anErrorHasOccurred)
+            } else {
+                val destinationDriveId = data.getIntExtra(DESTINATION_DRIVE_ID_KEY, -1)
+                val destinationFolderId = data.getIntExtra(DESTINATION_FOLDER_ID_KEY, -1)
```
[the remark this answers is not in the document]
FabianDevel: you're right, no need to make the api call as it would fail

app/src/main/java/com/infomaniak/drive/views/FileInfoActionsView.kt
```diff
@@ -609,5 +603,13 @@ class FileInfoActionsView @JvmOverloads constructor(
 
     companion object {
         const val SINGLE_OPERATION_CUSTOM_TAG = "single_operation"
+
+        fun Context.downloadFile(drivePermissions: DrivePermissions, file: File, onSuccess: (() -> Unit)? = null) {
```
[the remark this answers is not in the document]
FabianDevel: I put it in the interface's companion object, as it's here that it should belongs

tevincent (approved):

tevincent (approved):

app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
```diff
@@ -382,7 +392,7 @@ class SaveExternalFilesActivity : BaseActivity() {
     }
 
     private fun isValidFields(): Boolean {
-        return (isMultiple || !binding.fileNameEdit.showOrHideEmptyError()) &&
+        return (isMultiple || !binding.fileNameEdit.showOrHideEmptyError() || navigationArgs.isPublicShare) &&
```
sirambd, line 395: Why not just keep the same condition as before, since you're supposed to have a name every time, and so the name field should be completed?

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedListFragment.kt
```diff
@@ -122,8 +162,50 @@ class FileSharedListFragment : FileListFragment() {
         )
     }
 
+    private fun downloadAllFiles() {
+        fileSharedViewModel.rootSharedFile.value?.let { file ->
```
sirambd, line 166: Add a little comment here to explain that here we are dealing with either a file or a folder of files

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedListFragment.kt
```diff
@@ -122,8 +162,50 @@ class FileSharedListFragment : FileListFragment() {
...
+                fileIds = multiSelectManager.selectedItemsIds.toList(),
+                exceptedFileIds = multiSelectManager.exceptedItemsIds,
+            )
+        }
+    }
+
+    private fun setMainButton(importButton: MaterialButton) {
+        importButton.setOnClickListener {
+            if (AccountUtils.currentDriveId == -1) {
+                showSnackbar(title = "TODO : Show bottomsheet to get app", anchor = importButton)
```
sirambd, line 190: Fix this todo

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedListFragment.kt
```diff
@@ -122,8 +162,50 @@ class FileSharedListFragment : FileListFragment() {
...
+                destinationFolderId = destinationFolderId,
+                fileIds = multiSelectManager.selectedItemsIds.toList(),
+                exceptedFileIds = multiSelectManager.exceptedItemsIds,
+            )
+        }
+    }
+
+    private fun setMainButton(importButton: MaterialButton) {
+        importButton.setOnClickListener {
+            if (AccountUtils.currentDriveId == -1) {
```
sirambd, line 189: Does this really happen?

sirambd (changes requested):

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedListFragment.kt
FabianDevel: Yes if you open the link without being connected, you don't have a drive

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedListFragment.kt
FabianDevel: I won't dev a full new Ui bottomsheet in this PR because as the back isn't ready yet, we don't even know if this will be kept in the future.
So either we'll remove this button or i'll dev the bottomsheet.
What I can do however is removing the "showSnackbar" to be sure it's never displayed even if we forget about it

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedActivity.kt
```diff
@@ -43,21 +44,25 @@ class FileSharedActivity : AppCompatActivity() {
...
-            else -> {
+        val isMainButtonVisible = when (destination.id) {
+            R.id.fileSharedListFragment, R.id.fileSharedBottomSheetFileActions -> {
                 setColorStatusBar()
                 setColorNavigationBar()
+                true
             }
+            else -> false
         }
+        binding.mainFileShareButton.isVisible = isMainButtonVisible
```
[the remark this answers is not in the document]
FabianDevel: Because the FileSharedFragment doesn't have its own view, it's the binding of FileListFragment. I prefer not complicating it more than needed.

app/src/main/java/com/infomaniak/drive/ui/SaveExternalFilesActivity.kt
FabianDevel: In fact, in the case of public share, this activity is only needed to chose a destination, not to really upload file.
Then it's directly the back that will import all the files of the share to the destination.
So we don't chose a file whose name could be displayed, and we cannot fetch the name from realm.

app/src/main/java/com/infomaniak/drive/ui/fileShared/FileSharedListFragment.kt
sirambd: Add a TODO and redirect to login

FabianDevel: After discussion, a quick behaviour to wait for the back is to redirect to the login

sirambd (approved):

PR #1394, opened 2024-08-16
05 Add file action bottom sheet

FabianDevel:
Depends on #1393 

- Add a `PublicShareFileActionsBottomSheet` to execute actions on file without having to go to the preview
- Create an interface `OnPublicShareItemClicked` inheriting from `OnItemClicked` to factorize file action from the preview and the bottomsheet

PR #1395, opened 2024-08-16
06 Close the app and display a toast if the `init` route timed-out

FabianDevel:
Depends on #1394

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
@@ -42,9 +42,11 @@ import com.infomaniak.drive.utils.Utils
 import com.infomaniak.drive.utils.Utils.ROOT_ID
 import com.infomaniak.lib.applock.LockActivity
 import com.infomaniak.lib.applock.Utils.isKeyguardSecure
+import com.infomaniak.lib.core.api.ApiController
 import com.infomaniak.lib.core.extensions.setDefaultLocaleIfNeeded
 import com.infomaniak.lib.core.models.ApiResponseStatus
 import com.infomaniak.lib.core.utils.SentryLog
+import com.infomaniak.lib.core.utils.showToast
 import com.infomaniak.lib.stores.StoreUtils.checkUpdateIsRequired
 import io.sentry.Breadcrumb
 import io.sentry.Sentry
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
```diff
@@ -95,13 +98,18 @@ class LaunchActivity : AppCompatActivity() {
 
     private fun observeToFinishActivity() {
         canFinishActivity.observe(this@LaunchActivity) { canFinish ->
-            if (canFinish) {
-                lifecycleScope.launch {
+            if (!canFinish) return@observe
+
+            lifecycleScope.launch {
+                if (publicShareActivityExtras?.getBoolean(HAS_NO_NETWORK_TAG) == true) {
```
tevincent, line 104: Maybe we can simplify that part ?

tevincent (approved):

app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
```diff
@@ -188,6 +190,11 @@ class LaunchActivity : AppCompatActivity() {
                 }
                 ApiResponseStatus.REDIRECT -> apiResponse.uri?.let(::processInternalLink)
                 else -> {
+                    if (apiResponse.error?.exception is ApiController.NetworkException) {
+                        withContext(Dispatchers.Main) { showToast(R.string.errorNetwork) }
+                        finishAffinity()
+                        exitProcess(0)
```
sirambd, line 196: Why ?

app/src/main/AndroidManifest.xml
```diff
@@ -217,7 +217,8 @@
 
         <activity
             android:name=".ui.publicShare.PublicShareActivity"
-            android:configChanges="orientation|screenSize|layoutDirection" />
+            android:configChanges="orientation|screenSize|layoutDirection"
+            android:excludeFromRecents="true" />
```
sirambd, line 221: Why?

sirambd (changes requested):

app/src/main/AndroidManifest.xml
FabianDevel: Because I don't want this activity to make the app appearing in the "recent apps" on the user's phone
Just opening a link shouldn't count as a real use of the app

app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
FabianDevel: If the user has no network and can't load the share, we need to close the app because it failed to open.
The user can then try again to click on the link if he has a better connection

sirambd (approved):

PR #1402, opened 2024-08-27
07 Public Share fix multiselection

FabianDevel:
Depends on #1395 

Manage the multiselection for PublicShareFileList. The only supported action is the file's download.


tevincent (approved):

app/src/main/java/com/infomaniak/drive/ui/publicShare/PublicShareListFragment.kt
```diff
@@ -154,9 +165,11 @@ class PublicShareListFragment : FileListFragment() {
 
     private fun observeRootFile() {
         publicShareViewModel.rootSharedFile.observe(viewLifecycleOwner) { file ->
+            publicShareViewModel.fileClicked = file
```
sirambd, line 168: Is-it really useful ?

sirambd (changes requested):

app/src/main/java/com/infomaniak/drive/ui/publicShare/PublicShareListFragment.kt
FabianDevel: Yes otherwise the multiselection will crash with a NPE

sirambd (approved):

PR #1426, opened 2024-09-19
14 Move file url function to api route

FabianDevel:
Depends on #1425 

Move all the computing URL logic from File class to ApiRoutes with the rest of the URL

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/api/ApiRoutes.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/api/ApiRoutes.kt
@@ -22,6 +22,7 @@ import com.infomaniak.drive.data.api.UploadTask.Companion.ConflictOption
 import com.infomaniak.drive.data.models.File
 import com.infomaniak.drive.data.models.File.SortType
 import com.infomaniak.drive.utils.FileId
+import com.infomaniak.lib.core.BuildConfig
 import java.net.URLEncoder
 import java.util.Date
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/infomaniak/drive/data/api/ApiRoutes.kt
```diff
@@ -97,6 +98,45 @@ object ApiRoutes {
...
+    //region File url
+    fun getDownloadFileUrl(file: File): String = with(file) {
+        val downloadUrl = if (isPublicShared()) {
+            downloadPublicShareFile(driveId, publicShareUuid, id)
+        } else {
+            downloadFile(file)
+        }
+
+        return if (isOnlyOfficePreview()) "$downloadUrl?as=pdf" else downloadUrl
+    }
```
tevincent, line 111: What about doing extensions for these methods ?

For example:

```kotlin
fun File.getDownloadUrl(): String {
  val downloadUrl = if (isPublicShared()) {
      downloadPublicShareFile(driveId, publicShareUuid, id)
  } else {
      downloadFile(this)
  }

  return if (isOnlyOfficePreview()) "$downloadUrl?as=pdf" else downloadUrl
}
```

tevincent (changes requested):

tevincent:
Capture d’écran 2024-09-20 à 13 29 29

tevincent (approved):

tevincent (changes requested):
131

app/src/main/java/com/infomaniak/drive/data/api/ApiRoutes.kt
FabianDevel: If we put it as extension, we should put them in File.kt, that's a never ending question 😅

sirambd (approved):

tevincent (approved):

PR #1428, opened 2024-09-20
Fix upload of a file in a shared Drive

tevincent:
Not working:

https://github.com/user-attachments/assets/078a7824-16f1-4327-b868-21f78723b144

Working:

https://github.com/user-attachments/assets/8a043fd1-cad5-411e-8b2a-3f9334fc5377

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
@@ -57,22 +57,31 @@ class SelectFolderActivity : BaseActivity() {
         val customArgs = navigationArgs.customArgs
         val currentFolderId = navigationArgs.folderId.getIntOrNull()
         val disabledFolderId = navigationArgs.disabledFolderId.getIntOrNull()
-        val currentUserDrive = UserDrive(userId, driveId)
 
-        mainViewModel.selectFolderUserDrive = currentUserDrive
+        // We're doing this in the mainthread because the FileListFragment rely on mainViewModel.selectFolderUserDrive.
+        // Moving this call in a background thread we'll break everything
+        DriveInfosController.getDrive(driveId = driveId, maintenance = false)?.let { selectedDrive ->
+            val isSharedWithMe = selectedDrive.sharedWithMe
+            val currentUserDrive = UserDrive(userId, driveId, isSharedWithMe)
+            mainViewModel.selectFolderUserDrive = currentUserDrive
 
-        selectFolderViewModel.apply {
-            userDrive = currentUserDrive
-            currentDrive = DriveInfosController.getDrive(userId, driveId)
-            disableSelectedFolderId = disabledFolderId
-        }
+            selectFolderViewModel.apply {
+                userDrive = currentUserDrive
+                currentDrive = DriveInfosController.getDrive(userId, driveId)
+                disableSelectedFolderId = disabledFolderId
+            }
 
-        setSaveButton(customArgs)
+            setSaveButton(customArgs)
 
-        currentFolderId?.let { folderId ->
-            // Simply navigate when the folder exists in the local database
-            FileController.getFileProxyById(folderId, customRealm = mainViewModel.realm)?.let {
-                initiateNavigationToCurrentFolder(folderId, currentUserDrive)
+            currentFolderId?.let { folderId ->
+                // Simply navigate when the folder exists in the local database
+                FileController.getFileProxyById(
+                    fileId = folderId,
+                    userDrive = currentUserDrive,
+                    customRealm = getCustomRealm(currentUserDrive),
+                )?.let {
+                    initiateNavigationToCurrentFolder(folderId, currentUserDrive)
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
```diff
@@ -108,7 +116,7 @@ class SelectFolderActivity : BaseActivity() {
     }
 
     private fun MutableList<Int>.addNavigationIdsRecursively(folderId: Int, userDrive: UserDrive) {
-        FileController.getParentFileProxy(folderId, mainViewModel.realm)?.id?.let { parentId ->
+        FileController.getParentFile(folderId, userDrive, getRealm(userDrive))?.id?.let { parentId ->
```
FabianDevel, line 119: See with @sirambd  if we can use the getParentFile instead of the proxy here.

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
```diff
@@ -57,26 +57,34 @@ class SelectFolderActivity : BaseActivity() {
...
+                    userDrive = currentUserDrive,
+                    customRealm = getRealm(currentUserDrive)
+                )?.let {
+                    initiateNavigationToCurrentFolder(folderId, currentUserDrive)
+                }
             }
         }
     }
 
+    private fun getRealm(currentUserDrive: UserDrive) = if (currentUserDrive.sharedWithMe) null else mainViewModel.realm
```
FabianDevel, line 86: We should put that in a Utils somewhere, rename it to a more explicit name then use it in the other places where the same logic is used

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
```diff
@@ -57,26 +57,34 @@ class SelectFolderActivity : BaseActivity() {
...
-            FileController.getFileProxyById(folderId, customRealm = mainViewModel.realm)?.let {
-                initiateNavigationToCurrentFolder(folderId, currentUserDrive)
+            setSaveButton(customArgs)
+
+            currentFolderId?.let { folderId ->
+                // Simply navigate when the folder exists in the local database
+                FileController.getFileProxyById(
+                    fileId = folderId,
+                    userDrive = currentUserDrive,
+                    customRealm = getRealm(currentUserDrive)
```
FabianDevel, line 78:
```suggestion
                    customRealm = getRealm(currentUserDrive),
```

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
```diff
@@ -57,26 +57,34 @@ class SelectFolderActivity : BaseActivity() {
         val customArgs = navigationArgs.customArgs
         val currentFolderId = navigationArgs.folderId.getIntOrNull()
         val disabledFolderId = navigationArgs.disabledFolderId.getIntOrNull()
-        val currentUserDrive = UserDrive(userId, driveId)
+        DriveInfosController.getDrive(driveId = driveId, maintenance = false)?.let { currentDrive ->
```
FabianDevel, line 60: I think `selectedDrive`would suit better than `currentDrive`because it's the drive that is chosen by the user in the activity instead of the current drive on which he really is

FabianDevel: That would permit to remove the` this.` also

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
tevincent: I'll move it to the ViewModel and put the `realm` as private to avoid other person to use this realm without the proper parameters.

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
sirambd, line 60: Make it in background thread with a liveData

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
sirambd: Yes, it's a bit cumbersome to make a whole copy just to get an id, you have to go through a proxy.

sirambd (changes requested):

tevincent:
@sirambd Can you check where it makes sense to check for `sharedWithMe` ?

sirambd:
> @sirambd Can you check where it makes sense to check for `sharedWithMe` ?

Let's start with another PR, as it will be a global refactor

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
tevincent: For now, we'll fix this only here (see with @sirambd ). We'll check other places in another PR if needed.

app/src/main/java/com/infomaniak/drive/ui/fileList/SelectFolderActivity.kt
tevincent: As we discussed, we'll keep that here since moving that to a background thread will break everything (See comment).

FabianDevel (approved):

sirambd (approved):

PR #1441, opened 2024-09-30
fix: Syncing a file from an external app via cloud storage doesn't work

sirambd:
**Description**
When you open a file from an external application via cloud storage, the file is saved during editing, but when you close and open the file, it opens the old file in the cloud storage cache.

**Fixes**
- Ensure that file metadata is always up to date
- Manage cache to facilitate renewal

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
@@ -259,8 +259,7 @@ class CloudStorageProvider : DocumentsProvider() {
         val context = context ?: return null
 
         fun getRemoteFile(localFile: File?, fileId: Int, driveId: Int): File? {
-            val userId = getUserId(documentId)
-            val okHttpClient = runBlocking { AccountUtils.getHttpClient(userId.toInt()) }
+            val okHttpClient = runBlocking { AccountUtils.getHttpClient(getUserId(documentId).toInt()) }
             return ApiRepository.getFileDetails(localFile ?: File(id = fileId, driveId = driveId), okHttpClient).data
         }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
```diff
@@ -258,17 +258,29 @@ class CloudStorageProvider : DocumentsProvider() {
         SentryLog.d(TAG, "openDocument(), id=$documentId, mode=$mode, signalIsCancelled: ${signal?.isCanceled}")
         val context = context ?: return null
 
+        fun getRemoteFile(localFile: File?, fileId: Int, driveId: Int): File? {
+            val userId = getUserId(documentId)
+            val okHttpClient = runBlocking { AccountUtils.getHttpClient(userId.toInt()) }
+            return ApiRepository.getFileDetails(localFile ?: File(id = fileId, driveId = driveId), okHttpClient).data
+        }
+
```
tevincent, line 266:
```suggestion
    fun getRemoteFile(localFile: File?, fileId: Int, driveId: Int): File? {
         val okHttpClient = runBlocking { AccountUtils.getHttpClient(getUserId(documentId).toInt()) }
         return ApiRepository.getFileDetails(localFile ?: File(id = fileId, driveId = driveId), okHttpClient).data
    }
```
No need to create an extra variable for `userId`.

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
sirambd: Done

tevincent (approved):

PR #1445, opened 2024-10-01
Fix conflict between SwipeRefresh and FastScroller in gallery

TommyDL-Infomaniak:


app/src/main/java/com/infomaniak/drive/ui/menu/MenuGalleryFragment.kt
```diff
@@ -53,6 +54,14 @@ class MenuGalleryFragment : Fragment() {
         setUi(galleryFragment)
     }
 
+    fun initNoNetworkLayout(noNetworkInclude: LayoutNoNetworkSmallBinding) = with(binding) {
```
FabianDevel, line 57: put this method in GalleryFragment, and use "root" as the direct parent

FabianDevel (changes requested):

FabianDevel (approved):

PR #1463, opened 2024-10-09
Canceling small uploads works + doesn't popbackstack anymore

FabianDevel:
https://github.com/user-attachments/assets/74fb1aab-3bd9-4651-812d-3a597c473414

https://github.com/user-attachments/assets/91d290d6-3dfe-47f4-aa2b-8b8b14870367


sirambd (changes requested):
Make sure to dismiss the cancel upload dialog when upload fragment is destroyed

sirambd (approved):

PR #1477, opened 2024-10-25
Check before sending URIs to MediaStore that this one is not empty

tevincent:
We had a Sentry for a crash when we want to delete old uploaded files talking about some parcelable issue. 

I'm not entirely sure this fix is related to this Sentry but it'll surely do well.

The issue was that we could send a request to the MediaStore with an empty list of URIs that'll lead to a crash.


app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -366,29 +367,38 @@ class MainActivity : BaseActivity() {
...
+                            val filesDeletionRequest = MediaStore.createDeleteRequest(contentResolver, filesUriToDelete)
                             uploadedFilesToDelete = filesUploadedRecently
                             filesDeletionResult.launch(IntentSenderRequest.Builder(filesDeletionRequest.intentSender).build())
                         } else {
                             mainViewModel.deleteSynchronizedFilesOnDevice(filesUploadedRecently)
                         }
                     }
+
+                    // We check that the filtered list of URIs is not empty before showing the dialog and sending the request to
+                    // MediaStore; otherwise, it would cause a crash.
```
KevinBoulongne, line 391: I approve this PR, but I feel there is too much indentation :

image

KevinBoulongne: I've pushed another PR to enhance that : #1485.

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
KevinBoulongne: Result :

image

KevinBoulongne: Result :

image

KevinBoulongne: Result :

image

PR #1480, opened 2024-11-01
feat: Allow keyboard navigation

NicolasBourdin88:
Remove all keyboard traps and ensure that all interactive elements are focusable.


LunarX:
While navigating with the keyboard in order to activate the photo sync I had to select a folder. After a creating a new empty folder, I could not validate my folder choice with the keyboard, it seems to be focusing the background somehow instead. Am I missing an input?

image

LunarX:
Also when picking what folders of my phone to sync, my choice isn't taken into account when leaving the screen using the keyboard to click on the top left cross. I don't even get how this is possible

LunarX:
Also when previewing a file in the view pager, I can't seem to reach to top toolbar with the keyboard

LunarX:
When navigating inside folders and then navigating back up a level, the folder I just left is not selected anymore. Do you think it would be possible to select again the folder "we came from"?

LunarX (changes requested):
I've left the most important behaviors I've seen

NicolasBourdin88:
> When navigating inside folders and then navigating back up a level, the folder I just left is not selected anymore. Do you think it would be possible to select again the folder "we came from"?

Summary:
After looking into a solution, it seems challenging because the approach we're using in Infomaniak Mail doesn’t work with RecyclerView due to **ID** and **tag** recomposition.

app/src/main/java/com/infomaniak/drive/views/PreviewHeaderView.kt
```diff
@@ -108,6 +108,7 @@ class PreviewHeaderView @JvmOverloads constructor(
 
     fun setPageNumberVisibility(isVisible: Boolean) {
         binding.pageNumberChip.isVisible = isVisible
+        binding.pageNumberChip.isFocusable = false
```
LunarX, line 111: Add a comment so we understand why this is not done inside xml directly maybe

app/src/main/res/layout/fragment_file_list.xml
```diff
@@ -225,7 +227,8 @@
             <androidx.core.widget.NestedScrollView
                 android:layout_width="match_parent"
                 android:layout_height="match_parent"
-                android:fillViewport="true">
+                android:fillViewport="true"
+                android:touchscreenBlocksFocus="true">
```
LunarX, line 231: touchscreenBlocksFocus is also needed for nested scroll views? Or maybe rather, it's needed on the collapsing toolbar layout?

app/src/main/res/layout/fragment_home.xml
```diff
@@ -30,18 +30,21 @@
...
             android:layout_width="match_parent"
-            android:layout_height="wrap_content">
+            android:layout_height="wrap_content"
+            android:touchscreenBlocksFocus="false">
 
             <com.google.android.material.appbar.CollapsingToolbarLayout
                 android:id="@+id/collapsingToolbarLayout"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
+                android:touchscreenBlocksFocus="false"
```
LunarX, line 40: Same question, is it needed on the collapsing toolbar here? Because you usually don't specify it on these views. Where exactly should this android:touchscreenBlocksFocus be placed?

app/src/main/res/layout/fragment_preview_slider.xml
```diff
@@ -36,6 +36,9 @@
             android:id="@+id/viewPager"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
+            android:focusable="false"
+            android:focusableInTouchMode="false"
+            android:touchscreenBlocksFocus="true"
```
LunarX, line 41: Do you need all 3?

LunarX (changes requested):
Also the toolbar of the fragment_ment_gallery is not accessible

LunarX (approved):

PR #1483, opened 2024-11-06
Code cleaning

KevinBoulongne:


app/src/main/java/com/infomaniak/drive/data/cache/FileMigration.kt
```diff
@@ -255,7 +255,7 @@ class FileMigration : RealmMigration {
                 addRealmObjectField(File::conversion.name, fileConversionSchema)
                 addRealmObjectField(File::dropbox.name, dropboxSchema)
                 addRealmObjectField(File::version.name, fileVersionSchema)
-                addRealmObjectField(File::sharelink.name, shareLinkSchema)
+                addRealmObjectField(File::shareLink.name, shareLinkSchema)
```
FabianDevel, line 258: I wonder if you shouldn't add a migration when you rename a field, maybe check this

FabianDevel (approved):

PR #1487, opened 2024-11-11
Prevent crash when creating a new file

LunarX:
The app crashes for two users since this has been added. I prevented the crash and reported some more info to try and understand the issue.

Other places in the code call `createNewFile()` but I could not find crashes related to them in Sentry. Also, fixing them is not as straight forward as fixing it in this worker where we can easily return a `Failure`. For now, I did not touch any other places

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
+++ b/app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
@@ -46,6 +46,10 @@ import com.infomaniak.lib.core.models.ApiResponse
 import com.infomaniak.lib.core.networking.HttpClient
 import com.infomaniak.lib.core.networking.HttpUtils
 import com.infomaniak.lib.core.utils.SentryLog
+import io.sentry.Sentry
+import io.sentry.SentryEvent
+import io.sentry.SentryLevel
+import io.sentry.protocol.Message
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ensureActive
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
sirambd, line 1: Creates the parent folder if it doesn't exist

app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
```diff
@@ -170,8 +174,24 @@ class DownloadOfflineFileManager(
...
+        runCatching {
+            // This line is here to help some devices that don't succeed in automatically creating the file…
+            offlineFile.createNewFile()
+        }.onFailure {
+            Sentry.withScope { scope ->
+                val sentryEvent = SentryEvent().apply {
+                    level = SentryLevel.ERROR
+                    message = Message().apply { this.message = "Failed to create a new file" }
+                    throwable = it
+                }
```
sirambd, line 186: Why don't you just use a `Sentry.captureException`?

sirambd (changes requested):

app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
LunarX: I didn't know SentryLog.e could work with "extras". I changed it

PR #1498, opened 2024-12-03
feat: Remember the last file tree category

LouisCAD:
Depends on https://github.com/Infomaniak/android-core/pull/245


app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
```diff
@@ -82,46 +89,96 @@ class RootFilesFragment : Fragment() {
         )
 
         setupRootPendingFilesIndicator(mainViewModel.pendingUploadsCount, rootFilesUploadFileInProgressView)
+
+        if (fileListViewModel.justLaunched.not()) {
+            uiSettings.lastVisitedRootFileTreeCategory = null
+        }
+        if (fileListViewModel.justLaunched) lifecycleScope.launch {
```
FabianDevel, line 96: You should extract all the `lifecycleScope.launch` in its own function
The onCreateView method is already big, and this launch is really consequent

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
```diff
@@ -82,46 +89,96 @@ class RootFilesFragment : Fragment() {
...
+            fileListViewModel.justLaunched = false
+            when (uiSettings.lastVisitedRootFileTreeCategory) {
+                CommonFolders -> {
+                    folderToOpenSet.join()
+                    commonFolderToOpen?.let { (id, name) ->
+                        safeNavigate(RootFilesFragmentDirections.actionFilesFragmentToFileListFragment(folderId = id, folderName = name))
+                    }
+                }
+                PersonalFolder -> {
+                    folderToOpenSet.join()
```
FabianDevel, line 106: I dont really get what you were meaning with this name. 
What is something along the lines of 'setFolderToOpenJob` like it's the Job made for setting the folder to open ?

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
```diff
@@ -82,46 +89,96 @@ class RootFilesFragment : Fragment() {
...
+                    personalFolderToOpen?.let { (id, name) ->
+                        safeNavigate(RootFilesFragmentDirections.actionFilesFragmentToFileListFragment(folderId = id, folderName = name))
+                    }
+                }
+                Favorites -> {
+                    safeNavigate(RootFilesFragmentDirections.actionFilesFragmentToFavoritesFragment())
+                }
```
FabianDevel, line 113: We online the when entry when it's possible

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
FabianDevel: Also it will permit you to remove the null clause of the `when`because the compilator won't think you return something

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
LouisCAD: The `folderToOpenSet` property is a `CompletableJob`, a signal to await for the `commonFolderToOpen` and `personalFolderToOpen` properties to be set, since those are required to navigate to the "Common folder" and "Personal folder" routes.

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
```diff
@@ -82,46 +89,96 @@ class RootFilesFragment : Fragment() {
         )
 
         setupRootPendingFilesIndicator(mainViewModel.pendingUploadsCount, rootFilesUploadFileInProgressView)
+
+        if (fileListViewModel.justLaunched.not()) {
```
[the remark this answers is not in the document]
LouisCAD: Sounds like a habit thing rather than a clarity thing to me: `.not()` is more visible since it's a bit longer, and it's just next to the boolean expression it inverts, meaning you don't need to remember about the little `!` as one keeps reading the line.

app/src/main/java/com/infomaniak/drive/ui/home/RootFileTreeCategory.kt
```diff
@@ -0,0 +1,32 @@
...
+    // Don't remove entries, only rename and/or deprecate them.
+    // New entries must be added at the end.
+    CommonFolders,
+    PersonalFolder,
+    Favorites,
+    RecentChanges,
+    SharedWithMe,
+    MyShares,
+    Offline,
+    Trash;
```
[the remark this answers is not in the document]
LouisCAD: YOU_MEAN_LIKE_THIS? 😄 

From [Kotlin's official coding conventions](https://kotlinlang.org/docs/coding-conventions.html#property-names) (emphasis mine):

> For enum constants, it's OK to use either all uppercase, underscore-separated ([screaming snake case](https://en.wikipedia.org/wiki/Snake_case)) names (enum class Color { RED, GREEN }) **or upper camel case names**, depending on the usage.

I prefer "NonScreaming" for 3 reasons:
1. It is not putting unbalanced emphasis, helping reading
2. It is more common in reference Kotlin projects
3. It It aligns with the case used in other kinds of symbols that are used for use cases similar to enum entries: objects in sealed classes & interfaces, constants…

Any chance we can start adopting UpperCamelCase instead of UPPER_SNAKE_CASE for new code, starting with this PR? 🙏

app/src/main/java/com/infomaniak/drive/ui/fileList/FileListViewModel.kt
```diff
@@ -44,6 +44,8 @@ class FileListViewModel(application: Application) : AndroidViewModel(application
 
     private val realm = FileController.getRealmInstance()
 
+    var justLaunched = true
```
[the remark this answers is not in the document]
LouisCAD: Agreed, I'll add some doc to clarify the reason to be of this variable.

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
LouisCAD: > Also it will permit you to remove the null clause of the whenbecause the compilator won't think you return something

The compiler no longer tolerates non exhaustive when expressions when the subject type is an enum, a sealed class, or a sealed interface, so I'll need to keep that when branch.

app/src/main/java/com/infomaniak/drive/ui/home/RootFileTreeCategory.kt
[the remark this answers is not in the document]
FabianDevel: We'll talk about it in the team meeting to see what we want to do from now on

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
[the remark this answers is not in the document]
FabianDevel: It is a habit that define clarity ^^
I personnally find it way more clear with the `!`, probably because of the habit 
We'll discuss it also in meeting

app/src/main/java/com/infomaniak/drive/ui/home/RootFilesFragment.kt
LouisCAD: Thank you, I extracted all that logic to a new function. I agree it's more digest that way.

FabianDevel (approved):

sirambd (approved):

PR #1499, opened 2024-12-05
feat: Lock the app when the screen turns off

LouisCAD:
Depends on https://github.com/Infomaniak/android-core/pull/246


sirambd (changes requested):
It's worth testing, as it doesn't seem to work all the time.

LouisCAD:
Did some more testing, this should work reliably now.
One gotcha: If the phone is unlocked within 700ms after having been locked, the app lock might not trigger because of our usage of `ProcessLifecycle` that has a slight delay.

sirambd (approved):

PR #1507, opened 2024-12-27
fix: Empty stats icon color

JorisBodin:


LunarX:
Wait what, we want to remove the color from these files in recent activities?

From: 
Capture d’écran 2025-01-03 à 08 30 45 Capture d’écran 2025-01-03 à 08 30 55

To:
Capture d’écran 2025-01-03 à 08 31 51 Capture d’écran 2025-01-03 à 08 31 31

LunarX (approved):

PR #1526, opened 2025-02-19
fix: The Mediastore deletion request may generate a NullPointerException if the limit exceeds 10_000 uris.

sirambd:
- Make a chunk of 5000 per uris when we exceed, because even before 10_000 it happens that the query does not work.
- Avoid multiple dialogs when requesting media deletion
- Make sure that the request is made only in Start rather than in summary, which didn't make much sense.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
@@ -326,7 +326,7 @@ open class UploadFile(
         }
 
         private fun deleteFromRealm(realm: Realm, uploadFileUri: String) {
-            uploadFileByUriQuery(realm, uploadFileUri.toString()).findFirst()?.let { uploadFileRealm ->
+            uploadFileByUriQuery(realm, uploadFileUri).findFirst()?.let { uploadFileRealm ->
                 // Don't delete definitively if it's a sync
                 if (uploadFileRealm.type == Type.SYNC.name) {
                     uploadFileRealm.deletedAt = Date()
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -118,8 +119,15 @@ class MainActivity : BaseActivity() {
 
     private lateinit var drivePermissions: DrivePermissions
 
+    private var deleteLocalMediaRequestDialog: Dialog? = null
+    private val pendingFilesUrisQueue = ArrayDeque<List<Uri>>()
+
     private val filesDeletionResult = registerForActivityResult(StartIntentSenderForResult()) {
-        it.whenResultIsOk { lifecycleScope.launch(Dispatchers.IO) { UploadFile.deleteAll(uploadedFilesToDelete) } }
```
KevinBoulongne, line 122: The now unused `uploadedFilesToDelete` can now be removed.

KevinBoulongne (changes requested):

tevincent (approved):

app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
```diff
@@ -308,20 +308,34 @@ open class UploadFile(
...
+                it.executeTransaction { mutableRealm ->
+                    uris.forEach { uri ->
+                        deleteFromRealm(mutableRealm, uri.toString())
+                    }
+                }
+            }
+        }
+
+        private fun deleteFromRealm(realm: Realm, uploadFileUri: String) {
+            uploadFileByUriQuery(realm, uploadFileUri.toString()).findFirst()?.let { uploadFileRealm ->
```
KevinBoulongne, line 328:
```suggestion
            uploadFileByUriQuery(realm, uploadFileUri).findFirst()?.let { uploadFileRealm ->
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1546, opened 2025-02-28
feat(MatomoMyKSuite): Add my kSuite trackers

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/303


app/src/main/java/com/infomaniak/drive/MatomoDrive.kt
```diff
@@ -85,4 +86,16 @@ object MatomoDrive : MatomoCore {
...
         trackEvent("deepLink", name)
     }
+
+    fun Fragment.trackMyKSuiteEvent(name: String) {
+        trackEvent(MatomoMyKSuite.CATEGORY_MY_KSUITE, name)
+    }
+
+    fun Context.trackMyKSuiteEvent(name: String) {
+        trackEvent(MatomoMyKSuite.CATEGORY_MY_KSUITE, name)
+    }
```
tevincent, line 96: Is that necessary to have two different extensions, one for a `Fragment` and one for a `Context`? We can have the context in a `Fragment`.

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -272,6 +273,7 @@ class MainActivity : BaseActivity() {
     private fun setupFab(fab: FloatingActionButton, shouldShowSmallFab: Boolean = false) {
         val args = AddFileBottomSheetDialogArgs(shouldShowSmallFab).toBundle()
         fab.setOnClickListener {
+            trackMyKSuiteEvent("tryAddingFileWithDriveFull")
```
tevincent, line 276: Does this check need to be inside the `if (AccountUtils.getCurrentDrive()?.isDriveFull == true) { ...`?

Because it'll trigger even when the drive is not full right ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/utils/MyKSuiteNavUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import android.content.Context
+import androidx.fragment.app.Fragment
+import androidx.navigation.NavController
+import androidx.navigation.fragment.findNavController
+import com.infomaniak.core.myksuite.ui.screens.KSuiteApp
+import com.infomaniak.core.myksuite.ui.utils.MyKSuiteUiUtils.openMyKSuiteUpgradeBottomSheet
+import com.infomaniak.drive.MatomoDrive.trackMyKSuiteUpgradeBottomSheetEvent
+
+
```
KevinBoulongne, line 28:
```suggestion
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/drive/MatomoDrive.kt
KevinBoulongne: The Fragment one helps calling Matomo functions from fragments without needing to retrieve the context.
And the Context one allows calling Matomo functions from activities.

The 1st one is not mandatory, but a nice helper that we use in a lot of places.
The vast majority of Matomo functions are called from fragments.

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
FabianDevel: You're right of course, good catch

tevincent (approved):

PR #1550, opened 2025-03-06
feat(MyKSuite): Only fetch the data if the user has a kSuite

FabianDevel:


app/src/main/java/com/infomaniak/drive/utils/MyKSuiteDataUtils.kt
```diff
@@ -36,6 +36,11 @@ object MyKSuiteDataUtils : MyKSuiteDataManager() {
 
     override suspend fun fetchData(): MyKSuiteData? = runCatching {
         MyKSuiteDataUtils.requestKSuiteData()
+
+        // Doesn't try to fetch the kSuite Data is user doesn't have a kSuite
```
tevincent, line 40:
```suggestion
        // Don't try to fetch the kSuite Data if the user doesn't have a kSuite
```

tevincent (changes requested):

tevincent (approved):

PR #1555, opened 2025-03-13
fix: Drop too old schema 0 DriveInfos instead of crashing

LunarX:
DB that have configuration dating back to 4.4.7 or earlier that update to the 5.4.0 or superior keep on crashing when booting because the migration fails because it only supports DB configurations of 4.4.8+

This PR can detect if a migration failed due to the mentioned issue or not and delete the realm altogether as a migration in this only case. The reste of the time, the realm will continue crashing as usual if we ever make other mistakes in futur migrations

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
@@ -25,24 +25,23 @@ import com.infomaniak.drive.data.models.drive.Drive
 import com.infomaniak.drive.data.models.drive.DriveInfo
 import com.infomaniak.drive.utils.AccountUtils
 import com.infomaniak.drive.utils.RealmModules
-import io.realm.Realm
-import io.realm.RealmConfiguration
-import io.realm.RealmQuery
-import io.realm.Sort
+import io.realm.*
 import io.realm.kotlin.oneOf
 
 object DriveInfosController {
 
     private const val DB_NAME = "DrivesInfos.realm"
 
-    private val realmConfiguration = RealmConfiguration.Builder()
+    val baseDriveInfosRealmConfigurationBuilder: RealmConfiguration.Builder = RealmConfiguration.Builder()
         .name(DB_NAME)
         .schemaVersion(DriveMigration.DB_VERSION) // Must be bumped when the schema changes
-        .migration(DriveMigration())
         .modules(RealmModules.DriveFilesModule())
+
+    private val realmConfiguration = baseDriveInfosRealmConfigurationBuilder
+        .migration(DriveMigration())
         .build()
 
-    fun getRealmInstance(): Realm = Realm.getInstance(realmConfiguration)
+    fun getRealmInstance(): Realm = HandleSchemaVersionBelowZero.getInstance(realmConfiguration)
 
     private fun ArrayList<Drive>.initDriveForRealm(
         drive: Drive,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
```diff
@@ -25,24 +25,23 @@ import com.infomaniak.drive.data.models.drive.Drive
...
         .schemaVersion(DriveMigration.DB_VERSION) // Must be bumped when the schema changes
-        .migration(DriveMigration())
         .modules(RealmModules.DriveFilesModule())
+
+    private val realmConfiguration = baseDriveInfosRealmConfigurationBuilder
+        .migration(DriveMigration())
         .build()
 
-    fun getRealmInstance(): Realm = Realm.getInstance(realmConfiguration)
+    fun getRealmInstance(): Realm = HandleSchemaVersionBelowZero.getInstance(realmConfiguration)
```
FabianDevel, line 44: Maybe rename `getInstance()`to `getRealmInstance()`, otherwise we may think it's an instance of your `HandleSchemaVErsionBelowZeroObject` and that's really vague
Or maybe rename this object with `Realm` instead of` Object`

app/src/main/java/com/infomaniak/drive/data/cache/HandleSchemaVersionBelowZero.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.realm.Realm
+import io.realm.RealmConfiguration
+import io.realm.RealmMigration
+
+object HandleSchemaVersionBelowZero {
+    private val throwOldSchemaVersionRealmConfiguration = DriveInfosController.baseDriveInfosRealmConfigurationBuilder
+        .migration(DetectTooOldSchemaMigration())
+        .build()
+
+    private val deleteTooOldSchemaVersionRealmConfiguration = DriveInfosController.baseDriveInfosRealmConfigurationBuilder
```
FabianDevel, line 30: I think we can simplify this name, as it's really just a `deleteDataMigration`or `dropDataMigration`

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/data/cache/HandleSchemaVersionBelowZero.kt
sirambd, line 1: You can keep the same logic you have, but just make the `DriveMigration` class use a parameter that will only be used when it crashes, and in this case when you come from a schema version that is equal to 0 then you delete instead of migrating.

Here's an example:
https://github.com/Infomaniak/android-kDrive/blob/7d4c79861dda2c1656985a707b2a7ac842d8dbf2/app/src/main/java/com/infomaniak/drive/data/cache/FileMigration.kt#L392

sirambd (changes requested):

app/src/main/java/com/infomaniak/drive/data/cache/HandleSchemaVersionBelowZero.kt
LunarX: This doesn't work. It only removes all entries from the table, but the table's schema is still the same incompatible one

app/src/main/java/com/infomaniak/drive/data/cache/HandleSchemaVersionBelowZero.kt
LunarX: I took your idea and went with deleteRealmConfiguration to keep the same naming pattern

PR #1581, opened 2025-04-17
feat: Sync without right for all photo

FabianDevel:


app/src/main/java/com/infomaniak/drive/utils/DrivePermissions.kt
```diff
@@ -159,25 +166,30 @@ class DrivePermissions {
...
-            )
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> arrayOf(
-                Manifest.permission.ACCESS_MEDIA_LOCATION,
-                Manifest.permission.WRITE_EXTERNAL_STORAGE
-            )
-            else -> arrayOf(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+
+        @StringRes
+        val permissionNeededDescriptionRes = when {
+            SDK_INT >= VERSION_CODES.TIRAMISU -> R.string.allPermissionNeededAndroid13
```
tevincent, line 172: Do we generalize the use of API version value instead of the constant name? Because here, we check the API version by two different methods.

app/src/main/java/com/infomaniak/drive/utils/DrivePermissions.kt
```diff
@@ -159,25 +166,30 @@ class DrivePermissions {
...
+            else -> R.string.allPermissionNeeded
         }
 
-        fun Activity.resultPermissions(authorized: Boolean, permissions: Array<String>, message: Int) {
+
+        val permissions = buildSet {
+            if (SDK_INT in BuildConfig.MIN_SDK..<33) add(WRITE_EXTERNAL_STORAGE)
```
tevincent, line 178: On this: https://developer.android.com/reference/android/Manifest.permission#WRITE_EXTERNAL_STORAGE, it says if we target above API30, the permission is ignored. SO we should only add this permission if `SDK_INT in BuildConfig.MIN_SDK..<31` ? Or maybe I'm understanding the documentation the wrong way ?

app/src/main/java/com/infomaniak/drive/utils/DrivePermissions.kt
```diff
@@ -159,25 +166,30 @@ class DrivePermissions {
...
+            if (SDK_INT in 29..<CUR_DEVELOPMENT) add(ACCESS_MEDIA_LOCATION)
+            if (SDK_INT in 33..<CUR_DEVELOPMENT) {
```
tevincent, line 180: Just checked the `CUR_DEVELOPMENT` value and it's a magic number equals to `10000`

Capture d’écran 2025-04-22 à 17 27 38

Any reason to use that ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/utils/DrivePermissions.kt
FabianDevel: Yes, as it's stated in the doc it's the perfect const for this use case, where we want to target all current and future versions as long as the perm is not deprecated, as it always  represents the future build.

app/src/main/java/com/infomaniak/drive/utils/DrivePermissions.kt
FabianDevel: Probably I guess ? I just transcribed what was used before to avoid any unwanted bugs, but it indeed seems that 31 would be enough

app/src/main/java/com/infomaniak/drive/utils/DrivePermissions.kt
FabianDevel: I don't really know what we want to do, we always used constant names, but during the code'survey it was choosed for this place to use int values to keep short lines 🤷 
Anyway I guess it's a question for another refactor PR if we want to replace everything

app/src/main/java/com/infomaniak/drive/utils/DrivePermissions.kt
FabianDevel: It is needed for READ_EXTERNAL_STORAGE, which is needed up to API 32

tevincent (approved):

PR #1588, opened 2025-04-29
fix: Use translateError to avoid crashes

tevincent:
depends on https://github.com/Infomaniak/android-core/pull/340


LunarX (changes requested):

LouisCAD (approved):

app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
```diff
@@ -128,9 +129,9 @@ class DownloadOfflineFileManager(
         } else {
             if (fileDetails.error?.exception is ApiController.NetworkException) throw UploadTask.NetworkException()
 
-            val translatedError = fileDetails.translatedError
+            val translateError = fileDetails.translateError()
             val responseGsonType = object : TypeToken<ApiResponse<File>>() {}.type
-            val translatedErrorText = if (translatedError == 0) "" else context.getString(translatedError)
+            val translatedErrorText = if (translateError == 0) "" else context.getString(translateError)
```
LunarX, line 134: `translateError` can't be 0 anymore now that we use `translateError()`

LunarX (approved):

PR #1589, opened 2025-04-29
fix: Call the start method on Animatable drawables in order for GIF to start playing

tevincent:
It was working before because we were using the `imageView.load` but because we're handling the loading of images manually in order to hide/show views accordingly, we had to animate the image (like for a GIF) if it's a drawable of type Animatable.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPictureFragment.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPictureFragment.kt
@@ -18,7 +18,7 @@
 package com.infomaniak.drive.ui.fileList.preview
 
 import android.annotation.SuppressLint
-import android.graphics.drawable.Drawable
+import android.graphics.drawable.Animatable
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.MotionEvent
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPictureFragment.kt
```diff
@@ -85,18 +87,13 @@ class PreviewPictureFragment : PreviewFragment() {
...
-
         val imageLoader = Coil.imageLoader(requireContext())
         val thumbnailPreviewRequest = buildThumbnailPreviewRequest()
         val previewRequest = buildPreviewRequest()
 
         viewLifecycleOwner.lifecycleScope.launch(Dispatchers.IO) {
-            imageLoader.execute(thumbnailPreviewRequest).drawable?.let { setImageDrawable(it) }
-            imageLoader.execute(previewRequest).drawable?.let { setImageDrawable(it) }
+            imageLoader.execute(thumbnailPreviewRequest)
+            imageLoader.execute(previewRequest)
```
LunarX, line 96: Looks like we can use coil's `load()` method on the image view directly so we don't have to set the Drawable nor start the Animatable manually

app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPictureFragment.kt
```diff
@@ -85,18 +87,13 @@ class PreviewPictureFragment : PreviewFragment() {
     }
 
     private fun loadImage() = with(binding) {
```
LunarX, line 89: The way things are right now, `with(binding)` is unused

LunarX (changes requested):

KevinBoulongne (approved):

tevincent:
Seen with @sirambd. We'll keep as my modification is. The load is working with an `ImageView` but not with a `TouchImageView`.

LunarX (approved):

PR #1593, opened 2025-05-01
perf: Share/cache DB accesses for file categories

LouisCAD:
~~Depends on https://github.com/Infomaniak/android-core/pull/346~~
~~Depends on https://github.com/Infomaniak/android-core/pull/347~~
~~Depends on #1588~~


LouisCAD:
Doesn't build because of breaking changes in https://github.com/Infomaniak/android-core/pull/340
Will be resolved after it's rebased on commits in this PR: https://github.com/Infomaniak/android-kDrive/pull/1588

sirambd (approved):

PR #1600, opened 2025-05-09
chore: Update app logo icons

LunarX:
I used webp instead of png for the reduced file size like in SwissTransfer.

It's possible to compare the old icon with the new icon even if the file changed format if you look at the intermediate commits.

Here is what the icons look like:
Default app icon screenshot

Dynamic theme app icon light screenshot

Dynamic theme app icon dark screenshot


FabianDevel (changes requested):
Check for evenOdd simplification

FabianDevel (approved):

PR #1619, opened 2025-06-02
fix: Sentry file-move error

Muuushuuu:
Changed 'SerializedName' to 'Api Enum' and added the getAction() function to retrieve the correct action based on the received apiString

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/models/FileAction.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/models/FileAction.kt
@@ -19,7 +19,8 @@ package com.infomaniak.drive.data.models
 
 import android.os.Parcelable
 import com.google.gson.annotations.SerializedName
-import com.infomaniak.core.utils.enumValueOfOrNull
+import com.infomaniak.core.utils.apiEnum
+import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 
 @Parcelize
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/drive/data/models/FileAction.kt
```diff
@@ -32,5 +31,12 @@ data class FileAction(
...
 ) : Parcelable {
-    val actionType get() = enumValueOfOrNull<FileActivityType>(actionString.uppercase()) ?: FileActivityType.UNKNOWN
+    val actionType get() = getAction(actionString)
 }
+
+
+private fun getAction(actionString: String) : FileActivityType? {
+    return FileActivityType.entries.find(predicate = {
+        it.apiValue == actionString
+    })
```
LunarX: No need to be so verbose, put the lambda outside the parenthesis

```suggestion
    return FileActivityType.entries.find { it.apiValue == actionString }
```

Core
LunarX, line 1: No need for this Core update

LunarX (changes requested):

app/src/main/java/com/infomaniak/drive/data/models/FileAction.kt
Muuushuuu: it's done

Core
Muuushuuu: it's done

PR #1640, opened 2025-06-11
Add view with recent folders to root when selecting a folder

Muuushuuu:
This PR implements the display of a new section dedicated to recently modified folders in the root folder selection view.

When a user wishes to move or import a file, and reaches the root level of the selection, a new page is displayed, inspired by the existing iOS implementation.

This page displays : 
 - Classic root folders.
 - Up to 3 recently modified folders, grouped in a separate section at the top of the view.

The number of recent folders is configurable via a constant defined at the end of the relevant file, enabling rapid adjustment if required.

The design and behavior follow the indications of the Figma linked to the Trello card.

The view can be observed by :
 - Trying to move a file and navigating to the root.
 - Importing a file and accessing the destination folder selection.

PR #1643, opened 2025-06-12
refactor: Abstract classes compose with same code in 2 root files

Muuushuuu:
Depence on #1640

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/BaseRootFolder.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/BaseRootFolder.kt
@@ -1,0 +1,107 @@
+/*
+ * Infomaniak kDrive - Android
+ * Copyright (C) 2025 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.drive.ui
+
+import androidx.fragment.app.Fragment
+import androidx.navigation.NavDirections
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/drive/ui/BaseRootFolder.kt
```diff
@@ -0,0 +1,107 @@
...
+import com.infomaniak.core.fragmentnavigation.safelyNavigate
+import com.infomaniak.drive.data.models.File
+import com.infomaniak.drive.data.models.UiSettings
+import com.infomaniak.drive.databinding.RootFolderLayoutBinding
+import com.infomaniak.drive.ui.home.RootFileTreeCategory
+import com.infomaniak.drive.ui.home.RootFilesFragment.FolderToOpen
+import kotlinx.coroutines.CompletableJob
+import kotlinx.coroutines.Job
+
+abstract class BaseRootFolder : Fragment() {
```
tevincent: I would have called this `BaseRootFolderFragment` just to understand that this inherits from `Fragment`.

app/src/main/java/com/infomaniak/drive/ui/BaseRootFolder.kt
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            trashbin.setOnClickListener {
+                uiSettings.lastVisitedRootFileTreeCategory = RootFileTreeCategory.Trash
+                safelyNavigate(trashNav)
+            }
+        }
+    }
+
+    abstract fun fileListDirections(folderToOpen: FolderToOpen): NavDirections
```
tevincent: Group all the `abstract fun` together.

app/src/main/java/com/infomaniak/drive/ui/BaseRootFolder.kt
```diff
@@ -0,0 +1,107 @@
...
+        }
+        fileTypes[File.VisibilityType.IS_PRIVATE]?.let { file ->
+            personalFolderToOpen = FolderToOpen(file.id, file.getDisplayName(requireContext()))
+        }
+        if (haveBin) {
+            hasFolderToOpenBeenSet.complete()
+        }
+    }
+
+    abstract fun observeFiles()
```
tevincent: The observeFiles methods in `RootFIlesFragment` and `SelectRootFolderFragment` are almost the same. We could put the content of this method in `BaseRootFolder`?

Muuushuuu: the code is similar but the binding is not the same in both.

tevincent (approved):

PR #1648, opened 2025-06-17
feat: Custom view for the items in the settings

MorganeFar:


LunarX (changes requested):
Ok that's not what I meant when I said to do a first PR where you create the custom setting item view and replace it on the existing layout. We can talk about it when you're here

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+    fun setSwitchCheck(value: Boolean) {
+        binding.toggle.isChecked = value
+    }
+
+    private enum class Action {
+        NONE,
+        CHEVRON,
+        TOGGLE,
+        TEXT,
+    }
```
LunarX: Using SCREAMING_CASE to name enum entries is the old way of doing things, you can rename your enum using PascalCase

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    fun setTitle(title: String) {
+        binding.title.text = title
+    }
+
+    fun setDescription(description: String) {
+        binding.description.text = description
+    }
+
+    fun getTextEnd(): String {
```
LunarX: Maybe "end text" would more natural sounding than "text end"?

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+        binding.icon.imageTintList = ColorStateList.valueOf(color)
+    }
+
+    fun setIconEndVisibility(value: Boolean) {
+        binding.iconEnd.isVisible = value
+    }
+
+    fun setSwitchCheck(value: Boolean) {
+        binding.toggle.isChecked = value
+    }
```
LunarX: Name the arguments with more meaningful naming. `value` is a bit too vague for a situation like here where it's easy to find a more appropriate name for the argument

```suggestion
    fun setIconEndVisibility(isVisible: Boolean) {
        binding.iconEnd.isVisible = isVisible
    }

    fun setSwitchCheck(isChecked: Boolean) {
        binding.toggle.isChecked = isChecked
    }
```

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    fun setTextEnd(textEnd: String) {
+        binding.textEnd.text = textEnd
+    }
+
+    fun setOnCheckedChangeListener(listener: CompoundButton.OnCheckedChangeListener) {
+        binding.toggle.setOnCheckedChangeListener(listener)
+    }
+
+    fun setColorFolder(color: Int) {
```
LunarX:
```suggestion
    fun setColorFolder(@ColorInt color: Int) {
```

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+            binding.toggle.isChecked = value
+        }
+
+    init {
+        attrs?.getAttributes(context, R.styleable.ItemSettingView) {
+            with(binding) {
+                action = Action.entries[getInteger(R.styleable.ItemSettingView_itemAction, 0)]
+                title.text = getString(R.styleable.ItemSettingView_title) ?: ""
+                textEnd.text = getString(R.styleable.ItemSettingView_textEnd) ?: ""
+                description.text = getString(R.styleable.ItemSettingView_description) ?: ""
```
LunarX: A good practice is to avoid having two different ways of settings your values inside of your custom view.

Right now, when assigning a value during the initialization step, you access the view's property manually like in there with `title.text = getString(...` but then you also expose a `setTitle()` method for any other assignment of the value. This creates two "paths" in the code to do the same thing. 

This can be avoided if you directly call `setText()` when assigning a value during the initialization step. This way, you only have a single way of assigning these values in your whole code

app/src/main/java/com/infomaniak/drive/ui/menu/settings/SettingsFragment.kt
```diff
@@ -152,16 +152,16 @@ class SettingsFragment : Fragment() {
...
             .setNegativeButton(R.string.buttonCancel) { _, _ -> }
             .setCancelable(false).show()
     }
 
     override fun onResume() = with(binding) {
         super.onResume()
-        syncPictureValue.setText(if (AccountUtils.isEnableAppSync()) R.string.allActivated else R.string.allDisabled)
-        appSecurityValue.setText(if (AppSettings.appSecurityLock) R.string.allActivated else R.string.allDisabled)
+        syncPicture.setTextEnd(if (AccountUtils.isEnableAppSync()) getString(R.string.allActivated) else getString(R.string.allDisabled))
+        appSecurity.setTextEnd(if (AppSettings.appSecurityLock) getString(R.string.allActivated) else getString(R.string.allDisabled))
```
LunarX: You can extract the `getString()` part around the `if () ... else ...` part

```kt
getString(if (AccountUtils.isEnableAppSync()) R.string.allActivated else R.string.allDisabled)
```

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+        set(value) {
+            binding.toggle.isChecked = value
+        }
+
+    init {
+        attrs?.getAttributes(context, R.styleable.ItemSettingView) {
+            with(binding) {
+                action = Action.entries[getInteger(R.styleable.ItemSettingView_itemAction, 0)]
+                title.text = getString(R.styleable.ItemSettingView_title) ?: ""
+                textEnd.text = getString(R.styleable.ItemSettingView_textEnd) ?: ""
```
LunarX: There a nice trick you can use to write this title and textEnd properties without having to create a setTitle(), getTextEnd() or setTextEnd() and its to use property delegates.

You can expose two properties title and textEnd with which anyone can easily interact like this:

```kt
var title: CharSequence by binding.title::text
var textEnd: CharSequence by binding.textEnd::text
```

With this, no need to bother defining the getter and setter method by hand for these two because you don't manage any other form of logic

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+        toggle.setOnClickListener(listener)
+    }
+
+    fun setTitle(title: String) {
+        binding.title.text = title
+    }
+
+    fun setDescription(description: String) {
+        binding.description.text = description
+    }
```
LunarX: Because you are not using this method during initialization you did not realize that you missed computing the new visibility based on the new description's text. During the initialization step you set the visibility of the description based on if the the text is blank or not, but missed doing it here.

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+                title.text = getString(R.styleable.ItemSettingView_title) ?: ""
+                textEnd.text = getString(R.styleable.ItemSettingView_textEnd) ?: ""
+                description.text = getString(R.styleable.ItemSettingView_description) ?: ""
+
+                getDrawable(R.styleable.ItemSettingView_icon).let {
+                    icon.setImageDrawable(it)
+                    icon.isGone = it == null
+                }
+
+                description.isVisible = description.text.isNotBlank()
```
LunarX: To avoid negation you can actually write this one like this

```suggestion
                description.isGone = description.text.isBlank()
```

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,116 @@
...
+        binding.toggle.setOnCheckedChangeListener(listener)
+    }
+
+    fun setColorFolder(color: Int) {
+        binding.icon.imageTintList = ColorStateList.valueOf(color)
+    }
+
+    fun setIconEndVisibility(value: Boolean) {
+        binding.iconEnd.isVisible = value
+    }
```
LunarX: Maybe this could be generalized as changing from the NONE action to the TEXT action and vice versa?

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
LunarX: This setSwitchCheck() method is redundant with the isChecked property that you already exposed above

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
LunarX: This naming would be more grammatically correct

```suggestion
    fun setFolderColor(color: Int) {
```

app/src/main/res/layout/activity_sync_settings.xml
```diff
@@ -161,50 +142,14 @@
...
-                                    app:layout_constraintBottom_toBottomOf="parent"
-                                    app:layout_constraintEnd_toEndOf="parent"
-                                    app:layout_constraintTop_toTopOf="parent"
-                                    tools:ignore="ContentDescription"
-                                    tools:visibility="visible" />
-                            </androidx.constraintlayout.widget.ConstraintLayout>
+                                android:layout_height="wrap_content"
+                                android:minHeight="@dimen/buttonHeight"
+                                app:icon="@drawable/ic_drive"
+                                app:itemAction="none"
```
LunarX: Make it so "none" would be the default value if nothing is provided. It will make this custom view easier to use if you don't need to define "none" when you want no specific item action

app/src/main/res/layout/fragment_settings.xml
```diff
@@ -142,257 +119,89 @@
                     android:layout_height="match_parent"
                     android:orientation="vertical">
```
LunarX: You can actually remove all of the `View` that serves as dividers below as well by defining dividers at the `LinearLayout` level

```suggestion
                    android:divider="@drawable/divider"
                    android:showDividers="middle"
                    android:layout_height="match_parent"
                    android:orientation="vertical">
```

app/src/main/res/layout/view_item_setting.xml
```diff
@@ -0,0 +1,107 @@
...
+        app:layout_constraintEnd_toStartOf="@id/iconEnd"
+        app:layout_constraintStart_toEndOf="@id/icon"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_goneMarginStart="@dimen/marginStandard" />
+
+    <com.google.android.material.switchmaterial.SwitchMaterial
+        android:id="@+id/toggle"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="top"
```
LunarX: That line does nothing

```suggestion
```

app/src/main/res/layout/view_item_setting.xml
```diff
@@ -0,0 +1,107 @@
...
+        android:layout_marginEnd="@dimen/marginStandard"
+        android:paddingBottom="@dimen/marginStandardMedium"
+        android:text="@string/settingsOnlyWifiSyncDescription"
+        android:visibility="gone"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toStartOf="@id/iconEnd"
+        app:layout_constraintStart_toStartOf="@id/title"
+        app:layout_constraintTop_toBottomOf="@id/title"
+        app:layout_constraintVertical_chainStyle="packed"
+        tools:visibility="visible" />
```
LunarX: I feel like it would be more intuitive to define the description TextView right after the the title TextView as they are kind of similar instead of defining it after all of the end views with who they have much less in common

app/src/main/res/layout/view_item_setting.xml
```diff
@@ -0,0 +1,107 @@
...
+        android:layout_height="wrap_content"
+        android:layout_marginEnd="@dimen/marginStandard"
+        android:paddingBottom="@dimen/marginStandardMedium"
+        android:text="@string/settingsOnlyWifiSyncDescription"
+        android:visibility="gone"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toStartOf="@id/iconEnd"
+        app:layout_constraintStart_toStartOf="@id/title"
+        app:layout_constraintTop_toBottomOf="@id/title"
+        app:layout_constraintVertical_chainStyle="packed"
```
LunarX: This line has no effect the way it's used right there

```suggestion
```

app/src/main/res/values/attrs.xml
```diff
@@ -20,6 +20,8 @@
     <declare-styleable name="CommonAttributes">
         <attr name="icon" format="reference" />
         <attr name="title" format="string" />
+        <attr name="textEnd" format="string" />
+        <attr name="description" format="string" />
```
LunarX: No need to define these two up here if they are only used a single time. They are not common attributes used in multiple custom views

app/src/main/java/com/infomaniak/drive/ui/menu/settings/SyncSettingsActivity.kt
```diff
@@ -248,14 +246,13 @@ class SyncSettingsActivity : BaseActivity() {
...
-                driveName.text = it.name
+                selectDrive.setColorFolder(Color.parseColor(it.preferences.color))
+                selectDrive.setTitle(it.name)
                 selectDivider.isVisible = true
                 selectPath.isVisible = true
             } ?: run {
-                driveIcon.imageTintList =
-                    ColorStateList.valueOf(ContextCompat.getColor(this@SyncSettingsActivity, R.color.iconColor))
-                driveName.setText(R.string.selectDriveTitle)
+                selectDrive.setColorFolder(ContextCompat.getColor(this@SyncSettingsActivity, R.color.iconColor))
```
LunarX: Maybe the naming of `setColorFolder` is not appropriate and should be more generic and not mention "folder" directly because here it's not used to color a folder

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    fun setIconColor(@ColorInt color: Int) {
+        binding.icon.imageTintList = ColorStateList.valueOf(color)
+    }
+
+    fun setIconEndVisibility() {
+        binding.chevron.isVisible = action == Action.Chevron
+        binding.toggle.isVisible = action == Action.Toggle
+        binding.endText.isVisible = action == Action.Text
+    }
```
LunarX: There's 2 issues with this method.

First, the name of the method doesn't reflect what the method does. 

And second, you used to take an argument as input to change some values and I proposed to improve the way of doing so in https://github.com/Infomaniak/android-kDrive/pull/1648#discussion_r2165903933. Do you not need to change this value anymore? Because right now, this method will never have any impact.

We can talk about it directly

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,104 @@
...
+    var title: CharSequence by binding.title::text
+    var endText: CharSequence by binding.endText::text
+
+    init {
+        attrs?.getAttributes(context, R.styleable.ItemSettingView) {
+            with(binding) {
+                action = Action.entries[getInteger(R.styleable.ItemSettingView_itemAction, 0)]
+                title.text = getString(R.styleable.ItemSettingView_title) ?: ""
+                endText.text = getString(R.styleable.ItemSettingView_endText) ?: ""
+                setDescription(getString(R.styleable.ItemSettingView_description) ?: "")
```
LunarX: You still have the issue of https://github.com/Infomaniak/android-kDrive/pull/1648#discussion_r2163727753 in some of the lines here

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,104 @@
...
+            listener?.onClick(root)
+        }
+
+        toggle.setOnClickListener(listener)
+    }
+
+    fun setDescription(description: String) {
+        binding.description.text = description
+        binding.description.isGone = binding.description.text.isBlank()
+    }
```
LunarX: When there are redundant usages of a variable like here you can use a `with()`.

This could for exemple be rewritten as follows:
```suggestion
    fun setDescription(description: String) {
        with(binding.description) {
            text = description
            isGone = description.isBlank()
        }
    }
```

You can look `with()` up on the internet and then explain to me how it works

app/src/main/java/com/infomaniak/drive/ui/menu/settings/SyncSettingsActivity.kt
```diff
@@ -342,7 +337,7 @@ class SyncSettingsActivity : BaseActivity() {
     }
 
     private fun activeSelectDrive() = with(binding) {
-        switchDrive.isVisible = true
+        selectDrive.setIconEndVisibility()
```
LunarX: The code used to set the visibility to `true` but you stopped doing so. Is this really what we want and why?

app/src/main/java/com/infomaniak/drive/ui/menu/settings/SyncSettingsActivity.kt
```diff
@@ -378,8 +373,9 @@ class SyncSettingsActivity : BaseActivity() {
                 || allSyncedFoldersCount > 0
         saveButton.isVisible = isEdited
 
-        mediaFoldersTitle.text = if (allSyncedFoldersCount == 0) getString(R.string.noSelectMediaFolders)
+        val titleMediaFolders = if (allSyncedFoldersCount == 0) getString(R.string.noSelectMediaFolders)
         else resources.getQuantityString(R.plurals.mediaFoldersSelected, allSyncedFoldersCount, allSyncedFoldersCount)
+        mediaFolders.title = titleMediaFolders
```
LunarX: These lines could probably stay like they were before without the intermediate variable

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    override fun setOnClickListener(listener: OnClickListener?) = with(binding) {
+        root.setOnClickListener {
+            toggle.toggle()
+            listener?.onClick(root)
+        }
+
+        toggle.setOnClickListener(listener)
```
LunarX: Why do you need to modify the `toggle` item and add a click listener on it?

app/src/main/java/com/infomaniak/drive/ui/menu/settings/SyncSettingsActivity.kt
```diff
@@ -159,9 +158,8 @@ class SyncSettingsActivity : BaseActivity() {
         oldDeleteAfterSyncValue: Boolean,
     ) {
         setupSelectFolderListener()
-
-        activateSync.setOnClickListener { activateSyncSwitch.isChecked = !activateSyncSwitch.isChecked }
-        activateSyncSwitch.setOnCheckedChangeListener { _, isChecked ->
+        activateSync.setOnClickListener { activateSync.isChecked = !activateSync.isChecked }
+        activateSyncItem.setOnCheckedChangeListener { _, isChecked ->
```
LunarX: When clicking on the whole "Enable automatic backup" item, it doesn't toggle the item anymore compared to main. Only clicking on the small switch toggle works. Check if you can fix it to keep the behavior we have on main

LunarX (changes requested):

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -89,10 +87,8 @@ class ItemSettingView @JvmOverloads constructor(
         binding.icon.imageTintList = ColorStateList.valueOf(color)
     }
 
-    fun setIconEndVisibility() {
-        binding.chevron.isVisible = action == Action.Chevron
-        binding.toggle.isVisible = action == Action.Toggle
-        binding.endText.isVisible = action == Action.Text
+    fun setChevronVisibility(isVisible: Boolean) {
+        binding.chevron.isVisible = isVisible
```
LunarX: I think you still did not understand what I meant.

We could have a method `fun setAction(action: Action)`. When you want to hide the chevron you call it like this `setAction(Action.None)` and when you want to show the chevron you can call it like this `setAction(Action.Chevron)` instead of setting the visibility of the chevron specifically

app/src/main/java/com/infomaniak/drive/ui/menu/settings/ItemSettingView.kt
```diff
@@ -0,0 +1,100 @@
...
+
+            getDrawable(R.styleable.ItemSettingView_icon).let {
+                binding.icon.setImageDrawable(it)
+                binding.icon.isGone = it == null
+            }
+
+            binding.chevron.isVisible = action == Action.Chevron
+            binding.toggle.isVisible = action == Action.Toggle
+            binding.endText.isVisible = action == Action.Text
+
```
LunarX: You have one too many empty line

```suggestion
```

LunarX (changes requested):

LunarX (approved):

PR #1650, opened 2025-06-17
chore: Enhance usage of user current token

sirambd:


app/src/main/java/com/infomaniak/drive/MainApplication.kt
```diff
@@ -186,15 +188,26 @@ class MainApplication : Application(), ImageLoaderFactory, DefaultLifecycleObser
...
             AccountUtils.setUserToken(AccountUtils.currentUser!!, apiToken)
         }
 
         override suspend fun onRefreshTokenError() {
+            if (AccountUtils.currentUser == null) AccountUtils.requestCurrentUser()
             refreshTokenError(AccountUtils.currentUser!!)
         }
 
-        override suspend fun getApiToken(): ApiToken? = AccountUtils.currentUser?.apiToken
+        override suspend fun getApiToken(): ApiToken? = userTokenFlow.first()
```
LunarX:
```suggestion
        override suspend fun getApiToken(): ApiToken = userTokenFlow.first()
```

LunarX: Why only make sure that we will wait for a user token if not yet initialized in this TokenInterceptorListener but not in the one inside of CredentialManager?

app/src/main/java/com/infomaniak/drive/MainApplication.kt
```diff
@@ -186,15 +188,26 @@ class MainApplication : Application(), ImageLoaderFactory, DefaultLifecycleObser
     }
 
     private fun tokenInterceptorListener() = object : TokenInterceptorListener {
+        @OptIn(ExperimentalCoroutinesApi::class)
+        val userTokenFlow = AppSettings.getCurrentUserIdFlow()
+            .mapLatest { id -> id?.let { AccountUtils.getUserById(it)?.apiToken } }
+            .catch {
+                // Let user retry without logging him out.
+                // If we send null, the user will be logged out.
+            }
```
LunarX: Should we do this on mail as well?

app/src/main/java/com/infomaniak/drive/MainApplication.kt
```diff
@@ -186,15 +188,26 @@ class MainApplication : Application(), ImageLoaderFactory, DefaultLifecycleObser
...
+            .shareIn(applicationScope, SharingStarted.Lazily, replay = 1)
```
LunarX: There seems to be a lot of code in common with mail, should we not find a way to factorize some of it to avoid always doing everything twice or forgetting about the other project like with the `.catch` here?

app/src/main/java/com/infomaniak/drive/MainApplication.kt
sirambd: It can be null

app/src/main/java/com/infomaniak/drive/MainApplication.kt
sirambd: yes

app/src/main/java/com/infomaniak/drive/MainApplication.kt
sirambd: It's done.

LunarX (approved):

PR #1653, opened 2025-06-18
fix: Locat open folder 

Muuushuuu:
Depends on #1643


app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -89,6 +89,17 @@ object FileController {
         return realm?.let(block) ?: getRealmInstance(userDrive).use(block)
     }
 
+    fun getChildrenFileWithName(folderId: Int, name: String): List<Int> {
+        return getRealmInstance().use { realm ->
+             realm.where(File::class.java)
+                 .equalTo(File::parentId.name, folderId)
+                 .equalTo(File::name.name, name)
+                .equalTo(File::type.name, Type.DIRECTORY.value)
```
tevincent, line 97: Format your files.

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -89,6 +89,17 @@ object FileController {
         return realm?.let(block) ?: getRealmInstance(userDrive).use(block)
     }
 
+    fun getChildrenFileWithName(folderId: Int, name: String): List<Int> {
```
tevincent: You're getting files ids. The method of this method suggest we're getting the whole file.

app/src/main/java/com/infomaniak/drive/data/sync/UploadNotifications.kt
```diff
@@ -191,6 +192,17 @@ object UploadNotifications {
         )
     }
 
+    fun getDestinationFolderId(resources: UploadFile): Int {
+        val matches = resources.remoteSubFolder?.split("/")
+        var folderId: Int = resources.remoteFolder
+        if (matches != null) {
+            for (match in matches) {
+                folderId = FileController.getChildrenFileWithName(folderId, match)[0]
```
tevincent: It seems to crash on this line sometimes.

app/src/main/java/com/infomaniak/drive/data/sync/UploadNotifications.kt
```diff
@@ -191,6 +192,17 @@ object UploadNotifications {
         )
     }
 
+    fun getDestinationFolderId(resources: UploadFile): Int {
+        val matches = resources.remoteSubFolder?.split("/")
+        var folderId: Int = resources.remoteFolder
+        if (matches != null) {
+            for (match in matches) {
```
tevincent, line 199: Any reason to iterate through the whole path ? Because right now, we'll only get the last folder id in the path.

Could be something like this:

```kotlin
   fun getDestinationFolderId(resources: UploadFile): Int {
        var folderId = resources.remoteFolder

        resources.remoteSubFolder?.split("/")?.last()?.let { lastFolderInPath ->
            folderId = FileController.getChildrenFileWithName(folderId, lastFolderInPath)[0]
        }

        return folderId
    }
```

app/src/main/java/com/infomaniak/drive/ui/FileListNavigatoreObservable.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+package com.infomaniak.drive.ui
+
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.LifecycleOwner
+import com.infomaniak.drive.ui.fileList.FileListViewModel
+import com.infomaniak.drive.utils.FilePresenter.displayFile
+import com.infomaniak.drive.utils.FilePresenter.openFolder
+
+interface FileListNavigatoreObservable {
```
tevincent:
```suggestion
interface FileListNavigatorObservable {
```

app/src/main/java/com/infomaniak/drive/ui/FileListNavigatoreObservable.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    fun observeNavigateFileListTo(
```
tevincent: Any reason to put that here instead of `BaseRootFolderFragment` ? Also, the signature of this method is a bit overwhelming.

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
Muuushuuu: I only need the id to navigate to the right page (use in UploadNotifications L269) 
Then I don't know if it's better just to have the id or if I recover the folder and send folder.id

app/src/main/java/com/infomaniak/drive/data/sync/UploadNotifications.kt
Muuushuuu: is there a special case? I personally haven't had a crash.

app/src/main/java/com/infomaniak/drive/data/sync/UploadNotifications.kt
Muuushuuu: folder1 (id0) >folder2 (id1) >folder3 (id2)


yes, I think it's important, because we'll get the id of the father (folder1) and we'll get the string “folder2>folder3”.

but each time I'll check the name of the file and if its father corresponds to the idea I have

so here I'm checking if I have a “folder2” with parentId 0

and then I update parentId with 1

but if in folder3 we had a folder called “folder2” we wouldn't know what to take

app/src/main/java/com/infomaniak/drive/ui/FileListNavigatoreObservable.kt
Muuushuuu: I can't because my interface is for FileListFragment and RootFileFragment, but BaseRootFolderFragment extends only RootFileFragment. And for the complex signature how can I avoid providing them in the parameters ?

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
tevincent: No problem with that. What I'm saying is the method name is not what the method is really doing.

app/src/main/java/com/infomaniak/drive/data/sync/UploadNotifications.kt
tevincent: Just a simple upload. But what you need to check is if `getChildrenFileWithName` is empty to avoid any crash.

tevincent (approved):

PR #1654, opened 2025-06-19
fix deeplink from office (link)

Muuushuuu:


app/src/main/java/com/infomaniak/drive/data/models/UserDrive.kt
```diff
@@ -24,7 +24,7 @@ import kotlinx.parcelize.Parcelize
 @Parcelize
 data class UserDrive(
     val userId: Int = AccountUtils.currentUserId,
-    val driveId: Int = AccountUtils.currentDriveId,
+    var driveId: Int = AccountUtils.currentDriveId,
```
FabianDevel: Revert this

app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
```diff
@@ -227,7 +228,8 @@ class LaunchActivity : AppCompatActivity() {
     private fun setOpenSpecificFile(userId: Int, driveId: Int, fileId: Int, isSharedWithMe: Boolean) {
         if (userId != AccountUtils.currentUserId) AccountUtils.currentUserId = userId
         if (!isSharedWithMe && driveId != AccountUtils.currentDriveId) AccountUtils.currentDriveId = driveId
-        mainActivityExtras = MainActivityArgs(destinationFileId = fileId, isDestinationSharedWithMe = isSharedWithMe).toBundle()
+        val userDrive = UserDrive(sharedWithMe = isSharedWithMe, driveId = driveId)
+        mainActivityExtras = MainActivityArgs(destinationFileId = fileId, destinationUserDrive= userDrive).toBundle()
```
FabianDevel:
```suggestion
        mainActivityExtras = MainActivityArgs(destinationFileId = fileId, destinationUserDrive = userDrive).toBundle()
```

Try to always run the formatting after modifying a file

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -264,7 +264,7 @@ class MainActivity : BaseActivity() {
         navigationArgs?.let {
             if (it.destinationFileId > 0) {
                 clickOnBottomBarFolders()
-                mainViewModel.navigateFileListTo(navController, it.destinationFileId, it.isDestinationSharedWithMe)
+                mainViewModel.navigateFileListTo(navController, it.destinationFileId, it.destinationUserDrive!!)
```
FabianDevel: We try as much as possible to avoid the !! operator; 
There is little real use case
Here it's error prone just pass he nullability or define a default value

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -178,18 +178,16 @@ class MainViewModel(
             }
         }
     }
-
```
FabianDevel, line 206: revert this line

app/src/main/res/navigation/main_navigation.xml
```diff
@@ -342,9 +342,10 @@
             android:defaultValue="0"
             app:argType="integer" />
         <argument
-            android:name="isDestinationSharedWithMe"
-            android:defaultValue="false"
-            app:argType="boolean" />
+            android:name="destinationUserDrive"
+            app:argType="com.infomaniak.drive.data.models.UserDrive"
+            android:defaultValue="@null"
```
FabianDevel, line 347: This is not necessary to have the default value
And you can make it non nullable if you want to as you always pass it down

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
Muuushuuu: I replaced with `it.destinationUserDrive?.let` but I don't know if it's right or there's another way

app/src/main/res/navigation/main_navigation.xml
Muuushuuu: mainActivityArgs is used elsewhere without providing a userDrive (--> l.246 / p.LaunchActivity), so I don't always supply it

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -203,17 +203,16 @@ class MainViewModel(
...
-    fun navigateFileListTo(navController: NavController, fileId: Int, isSharedWithMe: Boolean = false) {
+    fun navigateFileListTo(navController: NavController, fileId: Int, userDrive: UserDrive) {
         // Clear FileListFragment stack
         navController.popBackStack(R.id.rootFilesFragment, false)
 
         if (fileId <= Utils.ROOT_ID) return // Deeplinks could lead us to navigating to the true root
 
         // Emit destination folder id
         viewModelScope.launch(Dispatchers.IO) {
-            val userDrive = UserDrive(sharedWithMe = isSharedWithMe)
```
FabianDevel, line 215: here the user drive was initiated with its default value for "driveId", so you should also use de default value in mainActivity instead of -1 (keeping the same behaviour is really important to avoid hard to track bug later ^^)

app/src/main/java/com/infomaniak/drive/ui/MainViewModel.kt
```diff
@@ -21,15 +21,7 @@ import android.app.Application
...
-import androidx.lifecycle.AndroidViewModel
-import androidx.lifecycle.LiveData
-import androidx.lifecycle.MediatorLiveData
-import androidx.lifecycle.MutableLiveData
-import androidx.lifecycle.SavedStateHandle
-import androidx.lifecycle.asLiveData
-import androidx.lifecycle.liveData
-import androidx.lifecycle.switchMap
-import androidx.lifecycle.viewModelScope
+import androidx.lifecycle.*
```
FabianDevel: We need to check your AS settings to avoid this
We don't want wildcard imports anymore (in order to follow android's guideline)

app/src/main/java/com/infomaniak/drive/utils/Extensions.kt
```diff
@@ -77,10 +77,7 @@ import com.infomaniak.drive.MatomoDrive.trackShareRightsEvent
 import com.infomaniak.drive.R
 import com.infomaniak.drive.data.api.ApiRoutes
 import com.infomaniak.drive.data.cache.DriveInfosController
-import com.infomaniak.drive.data.models.DriveUser
-import com.infomaniak.drive.data.models.File
-import com.infomaniak.drive.data.models.FileCategory
-import com.infomaniak.drive.data.models.Shareable
+import com.infomaniak.drive.data.models.*
```
FabianDevel: same here

FabianDevel (changes requested):

FabianDevel (approved):

PR #1657, opened 2025-06-19
feat: Add delete account system into settings 

Muuushuuu:


app/src/main/java/com/infomaniak/drive/ui/menu/settings/SettingsViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import android.content.Context
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.infomaniak.drive.utils.AccountUtils
+import com.infomaniak.lib.core.models.user.User
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.launch
+
+class SettingsViewModel(private val ioDispatcher: CoroutineDispatcher, private val appContext: Context): ViewModel() {
```
tevincent:
```suggestion
class SettingsViewModel(application: Application, private val ioDispatcher: CoroutineDispatcher): AndroidViewModel(application) {

    private inline val context: Context get() = getApplication<Application>().applicationContext
```

app/src/main/java/com/infomaniak/drive/ui/menu/settings/SettingsViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    fun disconnectDeletedUser(currentUser: User) {
+        viewModelScope.launch(ioDispatcher) {
+            AccountUtils.removeUserAndDeleteToken(appContext, currentUser)
+        }
+    }
+
```
tevincent:
```suggestion
```

app/src/main/java/com/infomaniak/drive/ui/menu/settings/SettingsViewModel.kt
tevincent: Usually, if you need a context inside a `ViewModel`, always use an `AndroidViewModel`.

tevincent (changes requested):

PR #1673, opened 2025-07-09
feat: Add UI to handle CrossAppLogin

KevinBoulongne:


app/build.gradle.kts
```diff
@@ -152,6 +152,7 @@ dependencies {
 
     implementation(project(":Core"))
     implementation(project(":Core:CrossAppLogin"))
+    implementation(project(":Core:CrossAppLoginUI"))
```
FabianDevel, line 155: Wouldn't a submodule inside CrossAppLogin would have more sense ? 
Like ":Core:CrossAppLogin:Ui" ?

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -184,10 +145,44 @@ class LoginActivity : AppCompatActivity() {
...
+        introViewpager.apply {
+            adapter = IntroPagerAdapter(supportFragmentManager, lifecycle)
+            registerOnPageChangeCallback(object : ViewPager2.OnPageChangeCallback() {
+                override fun onPageSelected(position: Int) {
+                    super.onPageSelected(position)
+                    val showConnectButton = position == 2
+                    nextButton.isGone = showConnectButton
+                    connectButton.isVisible = showConnectButton
+                    if (showConnectButton) {
+                        if (loginViewModel.crossLoginAccounts.value!!.isEmpty()) {
```
FabianDevel: Why not using `isNullOrEmpty()` ?

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -184,10 +145,44 @@ class LoginActivity : AppCompatActivity() {
...
+                            crossLoginSelection.isGone = true
+                            signInButton.isVisible = true
+                        } else {
+                            signInButton.isGone = true
+                            crossLoginSelection.isVisible = true
+                        }
+                    } else {
+                        crossLoginSelection.isGone = true
+                        signInButton.isGone = true
+                    }
```
FabianDevel: We could use a when here to only have one level of indentation
or extract the inner if to a function maybe ?

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -212,23 +207,137 @@ class LoginActivity : AppCompatActivity() {
...
+    }
+
+    private fun setCrossLoginClickListener() {
+
+        // Open Login webView when coming back from CrossLogin bottomSheet
+        supportFragmentManager.setFragmentResultListener(
+            /* requestKey = */ ON_ANOTHER_ACCOUNT_CLICKED_KEY,
+            /* lifecycleOwner = */ this,
+        ) { _, bundle ->
+            bundle.getString(ON_ANOTHER_ACCOUNT_CLICKED_KEY)?.let { openLoginWebView() }
```
FabianDevel, line 221: Shouldn't we return after this to avoid opening the bottomsheet under the webview ? 
If it works well, it seems like a side effect

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -212,23 +207,137 @@ class LoginActivity : AppCompatActivity() {
...
+    private fun initCrossLogin() = lifecycleScope.launch {
+        repeatOnLifecycle(Lifecycle.State.STARTED) {
+
+            loginViewModel.initDerivedTokenGenerator(coroutineScope = this)
+
+            val accounts = loginViewModel.getCrossLoginAccounts(context = this@LoginActivity)
+
+            if (accounts.isNotEmpty()) {
+                loginViewModel.crossLoginAccounts.value = accounts
+                loginViewModel.crossLoginSelectedIds.value = accounts.map { it.id }.toSet()
```
FabianDevel: What do you think about this : 
```suggestion
                loginViewModel.crossLoginSelectedIds.value = accounts.mapTo(mutableSetOf(), CrossLoginAccount::id)
```

or 
```suggestion
                loginViewModel.crossLoginSelectedIds.value = accounts.mapTo(mutableSetOf()) { it.id }
```

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -212,23 +207,137 @@ class LoginActivity : AppCompatActivity() {
...
+            }
+
+            repeatWhileActive {
+                binding.connectButton.awaitOneClick()
```
FabianDevel: In this awaitOneClick function, there is a useless explicit type arg

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -212,23 +207,137 @@ class LoginActivity : AppCompatActivity() {
...
+        }
+    }
+
+    private suspend fun handleCrossAppLogin() {
+
+        suspend fun authenticateToken(token: ApiToken, withRedirection: Boolean) {
+            authenticateUser(token, infomaniakLogin, withRedirection)
+        }
+
+        val accounts = loginViewModel.crossLoginAccounts.value
```
FabianDevel:
```suggestion
        val selectedAccounts = loginViewModel.crossLoginAccounts.value
```

?

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -212,23 +207,137 @@ class LoginActivity : AppCompatActivity() {
...
+            ?.filter { loginViewModel.crossLoginSelectedIds.value?.contains(it.id) == true }
+            ?: return
+        val tokenGenerator = loginViewModel.derivedTokenGenerator ?: return
```
FabianDevel: Shouldn't we handle the error cases instead of just returning ? 
Especially for the generator, is it "normal" that it could be null ?

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -212,23 +207,137 @@ class LoginActivity : AppCompatActivity() {
...
+
+        var firstAccountToken: ApiToken? = null
+        val accountsAndTokens = mutableMapOf<CrossLoginAccount, ApiToken?>()
+
+        accounts.forEach { account ->
+            val token = when (val result = tokenGenerator.attemptDerivingOneOfTheseTokens(account.tokens)) {
+                is Xor.First -> {
```
FabianDevel, line 271: Maybe add a commentary on first line of the first and secon d to explicit the types as it's "complicated" to get them (especially if you don't know what this xor is)

FabianDevel: Hmm the sentryLog kind of do this already though

FabianDevel (changes requested):

app/build.gradle.kts
KevinBoulongne: Will do !
But since there is multiple projects and around 10 currently open PRs about this module, I'll do it at the very end, just before merging everything into `main`.

PR #1683, opened 2025-07-17
fix: Account conflict 

Muuushuuu:
Depends on #1682 

To test, you need to be connected with 2 accounts.

- on account 1 you need to upload 1 file (to get a notification)
- disconnect from account 1 (switch to account 2)
- exit the app and click on the notification's locate button

there were conflicts between driveId and userId 
we were on account 2 with recent activity (drive) from account 1


app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
```diff
@@ -151,7 +153,21 @@ class LaunchActivity : AppCompatActivity() {
                     level = SentryLevel.INFO
                 })
 
-                setDeepLinkFileExtra(it.destinationDriveId, it.destinationRemoteFolderId)
+                CoroutineScope(Dispatchers.IO).launch {
```
LouisCAD: Never create a `CoroutineScope` that is not strongly referenced anywhere.
If the underlying coroutine is running on code that doesn't keep a strong reference to the cancellable continuation (and therefore the `Job` that is connected to parents and holds everything together), the coroutine could silently get garbage collected.

Here, you are in a `ComponentActivity` subclass, so you can use `lifecycleScope`.

To avoid unneeded blocking code, you could do something like this:

```kotlin
lifecycleScope.launch {
    val allUserIds = UserDatabase().userDao().allUsers.first().map { it.id }
    // Rest of the code, granted it's not doing blocking I/O
}
```

LouisCAD (changes requested):

app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
```diff
@@ -125,6 +125,18 @@ object DriveInfosController {
...
     }
 
+    fun deleteDriveIntoDriveInfo(userId: Int) {
+        getRealmInstance().use { realm ->
+            realm.executeTransaction {
+                val drives = realm.where(Drive::class.java)
+                    .equalTo(Drive::userId.name, userId)
+                    .findAll()
+
+                drives.map { it.deleteFromRealm() }
```
FabianDevel: `map` will return a list containing the result of what is in the lambda.
Here we don't want that, so just use a forEach : 

```suggestion
                drives.forEach(Drive::deleteFromRealm)
```

app/src/main/java/com/infomaniak/drive/data/cache/DriveInfosController.kt
```diff
@@ -125,6 +125,18 @@ object DriveInfosController {
         }
     }
 
+    fun deleteDriveIntoDriveInfo(userId: Int) {
```
FabianDevel: Just name it deleteDrive()

app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -466,6 +467,7 @@ object FileController {
                 }
             }
         }
+        deleteDriveIntoDriveInfo(userId)
```
FabianDevel: Here remove the existing import and write it like this : 
```suggestion
        DriveInfosController.deleteDrive(userId)
```

app/src/main/java/com/infomaniak/drive/ui/LaunchActivity.kt
```diff
@@ -151,7 +154,27 @@ class LaunchActivity : AppCompatActivity() {
...
                 })
 
-                setDeepLinkFileExtra(it.destinationDriveId, it.destinationRemoteFolderId)
+                lifecycleScope.launch {
+                    val allUserIds = UserDatabase.getDatabase()
+                        .userDao()
+                        .getAll()
+                        .asFlow()
+                        .first()
+                        .map { it.id }
```
FabianDevel:
```suggestion
                        .map(User::id)
```

FabianDevel (changes requested):

FabianDevel (approved):

LouisCAD (approved):

FabianDevel (approved):

PR #1696, opened 2025-07-24
fix: Use userDrive in geFileId function

Muuushuuu:


Muuushuuu:
Does not work correctly, the problem is deeper and is solved in `Folder name when save external file and select other drive #1702`.

PR #1698, opened 2025-07-24
feat: Cross applications login

KevinBoulongne:
Depends on https://github.com/Infomaniak/android-core/pull/449


app/src/main/java/com/infomaniak/drive/data/models/AppSettings.kt
```diff
@@ -71,14 +76,28 @@ open class AppSettings(
...
+            val flow = getRealmInstance()
+                .where(AppSettings::class.java)
+                .findFirst()
+                .toFlow()
+                .map { it?._currentUserId ?: -1 }
+            emitAll(flow)
+        }.flowOnLazyClosable {
+            @OptIn(DelicateCoroutinesApi::class, ExperimentalCoroutinesApi::class)
+            HandlerThreadDispatcher("Realm-currentUserIdFlow")
+        }
```
KevinBoulongne: Which one do we want to use ?

KevinBoulongne: The 2nd one, with some adjustments.

PR #1707, opened 2025-07-28
feat: Add cache and use cache for folder sync in photo backup

Muuushuuu:
When selecting a folder for photo synchronization, once all the photos had been deleted, it was no longer visible in `Folder to save`.
Added cache and used it to make the folder visible even if it was empty.


app/src/main/java/com/infomaniak/drive/ui/menu/settings/SelectMediaFoldersViewModel.kt
```diff
@@ -60,6 +60,26 @@ class SelectMediaFoldersViewModel : ViewModel() {
...
     }
 
+    private fun getLocalMediaFolders(
+        realm: Realm,
+        contentResolver: ContentResolver,
+        getMediaFilesJob: Job,
+        cacheMediaFolders: ArrayList<MediaFolder>,
+    ): ArrayList<MediaFolder> {
+
+        fun MediaFolder.exists(): Boolean = Environment.getExternalStoragePublicDirectory(path).exists()
```
FabianDevel: Use `appCtx.getExternalFilesDir(path)` instead of `Environment.getExternalStoragePublicDirectory(path)`

It's faster according to the documentation

FabianDevel (changes requested):

PR #1715, opened 2025-07-30
fix: Add runCatching to avoid a crash when adding an already existing file

tevincent:


app/src/main/java/com/infomaniak/drive/data/cache/FileController.kt
```diff
@@ -737,7 +737,9 @@ object FileController {
         // Save remote folder if it doesn't exist locally
         var newLocalFolderProxy: File? = null
         if (localFolderProxy == null && remoteFolder != null) {
-            realm.executeTransaction { newLocalFolderProxy = it.copyToRealm(remoteFolder) }
+            runCatching {
+                realm.executeTransaction { newLocalFolderProxy = it.copyToRealm(remoteFolder) }
+            }
```
FabianDevel: Maybe put a commentary to explain why we just catch everything without treating it

FabianDevel (approved):

PR #1722, opened 2025-07-31
feat: Fine grained permission requests

LouisCAD:
Instead of requesting almost all permissions, we now use a fine-grained approach, and consider optional permissions.

For example, the POST_NOTIFICATIONS permissions will be asked when relevant, but will no longer block the app from doing its job if not granted.


LouisCAD:
~~Found a regression in `UploadWorker` because I forgot to update several things in there. Will look it up tomorrow.~~

KevinBoulongne (approved):

PR #1741, opened 2025-08-13
fix: Avoid possible crash with out-of-scope Realm instance reuse

LouisCAD:


app/src/main/java/com/infomaniak/drive/data/cache/FolderFilesProvider.kt
```diff
@@ -52,14 +52,15 @@ object FolderFilesProvider {
 
     private val minDateToIgnoreCache = Calendar.getInstance().apply { add(Calendar.MONTH, -2) }.timeInMillis / 1000 // 3 month
 
-    fun getFiles(folderFilesProviderArgs: FolderFilesProviderArgs): FolderFilesProviderResult? {
-        val realm = folderFilesProviderArgs.realm ?: FileController.getRealmInstance(folderFilesProviderArgs.userDrive)
+    fun getFiles(
+        folderFilesProviderArgs: FolderFilesProviderArgs
+    ): FolderFilesProviderResult? = FileController.getRealmInstance(userDrive = folderFilesProviderArgs.userDrive).use { realm ->
```
sirambd: You haven't updated here, keep the realm test if defined from `folderFilesProviderArgs.realm`

sirambd (changes requested):

sirambd (approved):

PR #1742, opened 2025-08-13
fix: Update parent folder when refreshActivities is called

tevincent:
This is to always have an updated parent updatedAt with the right value.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/fileList/FileListFragment.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/fileList/FileListFragment.kt
@@ -590,8 +590,8 @@ open class FileListFragment : MultiSelectFragment(
         activitiesRefreshTimer.cancel()
         isLoadingActivities = true
         mainViewModel.currentFolder.value?.let { localCurrentFolder ->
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/ui/fileList/FileListFragment.kt
```diff
@@ -590,10 +590,10 @@ open class FileListFragment : MultiSelectFragment(
         activitiesRefreshTimer.cancel()
         isLoadingActivities = true
         mainViewModel.currentFolder.value?.let { localCurrentFolder ->
-            FileController.getFileById(localCurrentFolder.id, userDrive)?.let { updatedFolder ->
-                downloadFolderActivities(updatedFolder)
+            mainViewModel.getFileWithDetailsById(localCurrentFolder.id, userDrive) { updatedFolder ->
+                updatedFolder?.let { downloadFolderActivities(updatedFolder) }
                 activitiesRefreshTimer.start()
-            } ?: run { activitiesRefreshTimer.start() }
+            }
```
LunarX: You can avoid callbacks by making the viewmodel method suspend and using the lifecycle of the fragment

tevincent:
Won't fix since it seems to be an issue from the backend.

PR #1760, opened 2025-08-21
fix: Add marginEnd to prevent title sticking to edge

Muuushuuu:


FabianDevel (changes requested):
We should check all the other collapsing views to keep the same behavior

FabianDevel (approved):

PR #1777, opened 2025-09-05
fix: Edge to edge was not properly implemented for public shared files

tevincent:
Before fix:

https://github.com/user-attachments/assets/588eb2f9-2aaa-46e5-990f-fe251bf40287

After fix:

https://github.com/user-attachments/assets/cef06994-6ac4-4c3b-9d48-09abdaac80a9


app/src/main/java/com/infomaniak/drive/ui/publicShare/PublicShareActivity.kt
```diff
@@ -63,8 +65,12 @@ class PublicShareActivity : AppCompatActivity() {
         destination.trackDestination()
 
         if (destination.id == R.id.publicShareListFragment || destination.id == R.id.publicShareBottomSheetFileActions) {
+            binding.bottomNavigationBackgroundView.isGone = false
```
KevinBoulongne:
```suggestion
            binding.bottomNavigationBackgroundView.isVisible = true
```

PR #1778, opened 2025-09-05
fix: Improve generate thumbnail and getdata from cloudstorage

sirambd:
This is a refactor of certain methods that must be suspended to fix a Redmine issue on offline systems, but in passing it fixes and improves cloud storage, particularly on the `getData` and `generateThumbnail` methods.


app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
```diff
@@ -354,42 +358,55 @@ class CloudStorageProvider : DocumentsProvider() {
...
     }
 
-    private fun generateThumbnail(fileId: Int, file: File, userId: String): ParcelFileDescriptor? {
-
+    private fun generateThumbnail(fileId: Int, file: File, userId: String, signal: CancellationSignal?): ParcelFileDescriptor? {
         val outputFolder = IOFile(context?.cacheDir, "thumbnails").apply { if (!exists()) mkdirs() }
         val name = "${fileId}_${file.name}"
         val outputFile = IOFile(outputFolder, name)
-        var parcel: ParcelFileDescriptor? = null
+        val thumbnailUrl = ApiRoutes.getThumbnailUrl(file)
```
tevincent, line 373: Might be better to put this line near its use ? line 381.

app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
```diff
@@ -354,42 +358,55 @@ class CloudStorageProvider : DocumentsProvider() {
...
+        val job = cloudScope.launch(Dispatchers.IO) {
+            ParcelFileDescriptor.AutoCloseOutputStream(writePipe).use { writeStream ->
+                runCatching {
+                    val okHttpClient = AccountUtils.getHttpClient(userId.toInt())
+                    val response = DownloadOfflineFileManager.downloadFileResponseAsync(thumbnailUrl, okHttpClient)
 
-        } catch (exception: Exception) {
-            exception.printStackTrace()
+                    if (response.isSuccessful) {
+                        if (DownloadOfflineFileManager.saveRemoteData(TAG, response, outputFile)) {
```
tevincent: Any reason to have 2 `if`s instead of one ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
sirambd: It's an oversight, plus there's a case that isn't handled here.

tevincent (approved):

PR #1779, opened 2025-09-05
fix: Mark the file as offline only if it is saved on the disk  and improve sync offline stability

sirambd:
### Key Fixes:
- **Sync Migration**: Prevent common folders from being incorrectly removed during API v2 to v3 migration
- **Offline File Protection**: Fixed issues where offline files were unnecessarily deleted and ensured files are only marked as offline when successfully saved to disk
- **Permission Handling**: Added proper read permission checks before downloading files
- **Error Handling**: Improved download failure handling and prevented saving incomplete offline files
- **User Experience**: Reduced false notifications by only alerting users when actual errors occur

### Impact:
- More reliable offline file persistence
- Cleaner user notification experience with fewer false alerts
- Better permission security for file downloads
- Improved migration process stability

This work resolves several edge cases that were causing data loss and notification spam, resulting in a more robust offline file management system.

Depends on #1778

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/services/DownloadWorker.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/services/DownloadWorker.kt
@@ -50,9 +50,7 @@ class DownloadWorker(context: Context, workerParams: WorkerParameters) : BaseDow
     }
 
     override fun downloadNotification(): DownloadNotification? {
-        return file?.id?.let { notificationId ->
-            DownloadNotification(id = notificationId, notification = downloadProgressNotification)
-        }
+        return DownloadNotification(id = fileId, notification = downloadProgressNotification)
     }
 
     override suspend fun downloadAction(): Result = downloadFile()


app/src/main/java/com/infomaniak/drive/data/services/DownloadWorker.kt
```diff
@@ -86,6 +80,9 @@ class DownloadWorker(context: Context, workerParams: WorkerParameters) : BaseDow
 
             if (result == Result.success()) {
                 applicationContext.cancelNotification(fileId)
+            } else {
+                notifyDownloadCancelled()
+                file?.id?.let(notificationManagerCompat::cancel)
```
tevincent: Could be simplified like that right ?

```kotlin
            if (result == Result.failure()) notifyDownloadCancelled()
            file?.id?.let(notificationManagerCompat::cancel)
```

tevincent (changes requested):

PR #1795, opened 2025-09-22
chore: Ensure no double login can happen

LouisCAD:
Will allow removing `computeSelectedAccounts(…)` and `selectedAccounts` in `BaseCrossAppLoginViewModel` once the Mail app is updated.


app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -151,8 +148,18 @@ class LoginActivity : ComponentActivity() {
...
 
-    private suspend fun connectSelectedAccounts(accounts: List<ExternalAccount>, skippedIds: Set<Long>) {
-        val selectedAccounts = BaseCrossAppLoginViewModel.computeSelectedAccounts(accounts, skippedIds)
+    private suspend fun handleLogin(loginRequest: CallableState<List<ExternalAccount>>): Nothing = repeatWhileActive {
+        val accountsToLogin = loginRequest.awaitOneCall()
+        if (accountsToLogin.isEmpty()) {
+            openLoginWebView()
+        } else {
+            val skippedIds = crossAppLoginViewModel.skippedAccountIds.value
+            val selectedAccounts = accountsToLogin.filter { it.id !in skippedIds }
```
LunarX: `accountsToLogin` doesn't represent the list of accounts to login but the whole list of accounts. Maybe it should be renamed or the pair of all accounts and skipped ids should be passed to the loginRequest and we can destructure it here when we receive it.

Or we can pass in the already computed accounts to login when requesting login

LunarX: We should call the factorized code as long as the code is duplicated
```
val selectedAccounts = computeSelectedAccounts(accountsToLogin, skippedIds)
```

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -151,8 +148,18 @@ class LoginActivity : ComponentActivity() {
         initCrossLogin()
     }
 
-    private suspend fun connectSelectedAccounts(accounts: List<ExternalAccount>, skippedIds: Set<Long>) {
-        val selectedAccounts = BaseCrossAppLoginViewModel.computeSelectedAccounts(accounts, skippedIds)
+    private suspend fun handleLogin(loginRequest: CallableState<List<ExternalAccount>>): Nothing = repeatWhileActive {
```
LunarX, line 151: Maybe it would be even better to send a sealed class/interface instead of a list inside loginRequest

app/src/main/java/com/infomaniak/drive/ui/login/LoginActivity.kt
```diff
@@ -151,8 +148,18 @@ class LoginActivity : ComponentActivity() {
...
+        val accountsToLogin = loginRequest.awaitOneCall()
+        if (accountsToLogin.isEmpty()) {
+            openLoginWebView()
+        } else {
+            val skippedIds = crossAppLoginViewModel.skippedAccountIds.value
+            val selectedAccounts = accountsToLogin.filter { it.id !in skippedIds }
+            connectAccounts(selectedAccounts = selectedAccounts)
```
LunarX: Talking with @sirambd we thought about maybe extracting each branch of the if/else or the when (if we start using a sealed class) into its own function. This could be achieved by passing the new sealed class into the existing `connectAccounts()` method

PR #1797, opened 2025-09-23
chore: Refactor sentry to use inside scope

FabianDevel:
Since some versions, Sentry doesn't upload the extra when using `Sentry.withScope`. 
So we use the existing scope property from the existing `captureMessage` and `captureException` function.

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/models/File.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/models/File.kt
@@ -1,6 +1,6 @@
 /*
  * Infomaniak kDrive - Android
- * Copyright (C) 2022-2024 Infomaniak Network SA
+ * Copyright (C) 2022-2025 Infomaniak Network SA
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/drive/ui/bottomSheetDialogs/BackgroundSyncPermissionsBottomSheetDialog.kt
```diff
@@ -112,14 +112,11 @@ class BackgroundSyncPermissionsBottomSheetDialog : BottomSheetDialogFragment() {
                 Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS,
                 "package:$packageName".toUri()
             ).apply { permissionResultLauncher.launch(this) }
-        } catch (activityNotFoundException: ActivityNotFoundException) {
-            try {
+        } catch (_: ActivityNotFoundException) {
```
KevinBoulongne, line 115: You can also replace this `try / catch` with `runCatching / onFailure`.

PR #1799, opened 2025-09-23
fix: Preview of files in PublicShare when not connected now works

tevincent:


app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewSliderAdapter.kt
```diff
@@ -41,9 +42,9 @@ class PreviewSliderAdapter(
         val args = PreviewFragmentArgs(fileId = file.id, userDrive = userDrive).toBundle()
 
         return when (file.getFileType()) {
-            ExtensionType.IMAGE -> PreviewPictureFragment()
-            ExtensionType.VIDEO -> PreviewVideoFragment()
-            ExtensionType.AUDIO -> PreviewMusicFragment()
+            ExtensionType.IMAGE -> PreviewPictureFragment(isPublicShared)
+            ExtensionType.VIDEO -> PreviewVideoFragment(isPublicShared)
+            ExtensionType.AUDIO -> PreviewMusicFragment(isPublicShared)
```
sirambd: For fragments, don't pass parameters like that because it's dangerous and creates bugs after recreate. Instead, use `Fragment.arguments`.

sirambd (changes requested):

sirambd (approved):

PR #1802, opened 2025-09-25
refactor(PublicShareApi): Create a new api repository specific to publicShare

FabianDevel:
Use non blocking callApi in the same time

PR #1803, opened 2025-09-25
fix: Fix public share password 401

FabianDevel:
Depends on #1802 

Create new *HttpClient* specific to publicShare to make it agnostic from the change of the app. 
Override the new `BaseHttpClient` from the *Auth* module to make its own `TokenInterceptor`

Now the interceptor won't disconnect user if a authenticated call returns a 401, because there are "normal" flow cases based on this (password protected share and expired share).
Also, the token will be passed only if there are user connected, because public share can be used without connection, but we still need the token to have the redirection to a internal link if the connected user already has access to the shared file in its drive.

In a next PR I will refactor Current HttpClient to use the *Auth* module instead of the legacy one

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/api/publicshare/PublicShareApiRepository.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/api/publicshare/PublicShareApiRepository.kt
@@ -17,6 +17,7 @@
  */
 package com.infomaniak.drive.data.api.publicshare
 
+import com.infomaniak.core.network.networking.HttpClient
 import com.infomaniak.drive.data.api.ApiRoutes
 import com.infomaniak.drive.data.api.ApiRoutes.loadCursor
 import com.infomaniak.drive.data.api.CursorApiResponse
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/data/api/publicshare/PublicShareHttpClient.kt
```diff
@@ -0,0 +1,32 @@
...
+import okhttp3.OkHttpClient
+
+object PublicShareHttpClient : BaseHttpClient() {
+
+    override fun OkHttpClient.Builder.addTokenInterceptor() {
+
+        AccountUtils.currentUser?.let {
+            tokenInterceptorListener?.let { listener -> addInterceptor(PublicShareTokenInterceptor(listener)) }
+        }
+    }
```
KevinBoulongne, line 30:
```suggestion
    override fun OkHttpClient.Builder.addTokenInterceptor() {
        if (AccountUtils.currentUser == null) return

        tokenInterceptorListener?.let { listener -> addInterceptor(PublicShareTokenInterceptor(listener)) }
    }
```

app/src/main/java/com/infomaniak/drive/data/api/publicshare/PublicShareApiRepository.kt
sirambd, line 1: I'm waiting for the changes we talked about.

sirambd (changes requested):

sirambd (approved):

PR #1811, opened 2025-09-29
fix: Check if the api resume chunk size is correct before resuming the upload

sirambd:
Check the api expected chunk size and add more sentry log when an unexpected error has been occurred

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/api/UploadTask.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/api/UploadTask.kt
@@ -195,7 +195,7 @@ class UploadTask(
     }
 
     private fun getChunkConfig(validChunks: ValidChunks?): FileChunkSizeManager.ChunkConfig {
-        val validChuckSize = validChunks?.validChuckSize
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/drive/data/api/UploadTask.kt
```diff
@@ -195,12 +194,22 @@ class UploadTask(
...
     }
 
-    private fun getChunkConfig(): FileChunkSizeManager.ChunkConfig {
+    private fun getChunkConfig(validChunks: ValidChunks?): FileChunkSizeManager.ChunkConfig {
         return try {
             fileChunkSizeManager.computeChunkConfig(
                 fileSize = uploadFile.fileSize,
-                defaultFileChunkSize = uploadFile.getValidChunks()?.validChuckSize?.toLong(),
-            )
+                defaultFileChunkSize = validChunks?.validChuckSize?.toLong(),
```
FabianDevel:
```suggestion
                defaultFileChunkSize = validChunks?.validChunckSize?.toLong(),
```

PR #1817, opened 2025-10-01
fix(RemoveCaptureVideoIntent): Remove the choice to take a video in the add file bottomSheet

FabianDevel:
This was bugged on xiaomi devices, where the chosser wouldn't open correctly and always force photo


app/src/main/java/com/infomaniak/drive/ui/addFiles/AddFileBottomSheetDialog.kt
tevincent: Not used.

tevincent, line 157: Do we want to rename that Matomo ?

tevincent (changes requested):

tevincent (approved):

PR #1818, opened 2025-10-01
fix: Bottom navigation bar had too much bottom padding

tevincent:


app/src/main/java/com/infomaniak/drive/ui/BasePreviewSliderFragment.kt
```diff
@@ -249,7 +249,6 @@ abstract class BasePreviewSliderFragment : Fragment(), FileInfoActionsView.OnIte
 
     private fun clearEdgeToEdge() = with(requireActivity()) {
         toggleSystemBar(true)
-        window.toggleEdgeToEdge(false)
```
LunarX, line 252: You did not remove the import for this method in this file

PR #1822, opened 2025-10-02
fix: Change how we build fDroid on the CI to detect more issues

tevincent:


.github/workflows/android.yml
```diff
@@ -32,6 +34,8 @@ jobs:
 
       - name: Create test env
         run: |
+          touch env.properties
```
sirambd: Use the env SENTRY_AUTH_ENV_PATH instead

sirambd (approved):

PR #1829, opened 2025-10-06
fix: Handle Foreground quota exceed exception

tevincent:


app/src/main/res/values/strings.xml
```diff
@@ -766,4 +768,4 @@
     <string name="userPermissionRemove">User will no longer have access to this file.</string>
     <string name="userPermissionWrite">Can change</string>
     <string name="userPermissionWriteDescription">Edit the file\nDownload\nAdd comments\nAdd and create files/folders\nDeleting the file</string>
-</resources>
+</resources>
```
KevinBoulongne, line 771: The 5 strings files have lost their last line.

app/src/main/java/com/infomaniak/drive/data/services/UploadWorkerThrowable.kt
```diff
@@ -43,42 +46,43 @@ object UploadWorkerThrowable {
...
-            Result.retry()
-
-        } catch (exception: UploadTask.LockErrorException) {
+        } catch (_: CancellationException) {
+            // A CancellationException is thrown if, of course, the worker has been cancelled but also if the quota for foreground
+            // services has been reached. Since Android 15, app's foreground services is authorized to run 6 hours / day in the
+            // background at most. To reset this quota, the app need to be brought in foreground. So in that case, we display a
+            // notification to ask the user to go back in the app.
+            // See https://developer.android.com/develop/background-work/services/fgs/timeout for more info.
+            if (Build.VERSION.SDK_INT >= 31 && stopReason == STOP_REASON_FOREGROUND_SERVICE_TIMEOUT) {
```
FabianDevel, line 68: Why 31 ? if it's from android 15 it should be 35

PR #1837, opened 2025-10-08
chore: Bump material

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/541
Depends on #1839


app/src/main/java/com/infomaniak/drive/data/services/DownloadReceiver.kt
sirambd, line 1: Move to BaseDownloadWorker

PR #1842, opened 2025-10-09
fix: Fix CloudStorageProvider crash for drives with @ in their name

LouisCAD:
Couldn't test it because the backend doesn't let me change the drive name for the test account.

@sirambd Could you test it on your side?


app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
```diff
@@ -891,8 +891,13 @@ class CloudStorageProvider : DocumentsProvider() {
...
             return if (documentId.contains(DRIVE_SEPARATOR)) {
-                val drive = parentDocumentId.substringAfter(SEPARATOR).split(DRIVE_SEPARATOR)
-                DriveDocument(name = drive[0], id = drive[1].toInt())
+                val drivePath = parentDocumentId.substringAfter(SEPARATOR)
+                val driveName = drivePath.substringBeforeLast(DRIVE_SEPARATOR)
+                val driveId = drivePath.substringAfterLast(DRIVE_SEPARATOR).toInt()
+                DriveDocument(
+                    name = driveName,
+                    id = driveId
+                )
```
KevinBoulongne:
```suggestion
                DriveDocument(driveName, driveId)
```

KevinBoulongne:
Ok for me if @sirambd 's tests are OK.

PR #1873, opened 2025-11-11
Adding Russian metadata translation

Korb:
https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#in-the-applications-source-repository


JorisBodin:
Hello @Korb 

Thank you very much for your contribution

But, as explained [here](https://github.com/Infomaniak/android-SwissTransfer/pull/541#issuecomment-3312773837), we do not support languages other than the current five.

Maybe in the future this will change, and when it does, we will provide translations.

PR #1872, opened 2025-11-11
feat(TwoFactorAuth): Add 2FA notifications support

LouisCAD:
Depends on https://github.com/Infomaniak/android-core/pull/593

PR #1874, opened 2025-11-12
feat: Better cross app login UI

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/595
Depends on #1872
Closes #1877 

This PR adds a check of account's derivability before displaying it to the user. 
As this check relies on an api call, we needed to change the UI to make it smoother. 

There is now a loader between when the IPC returns the list of account, and when the first account is checked. 
Each accounts are now simultaneously checked, and displayed one by one as soon as they are checked. (Prior to this PR, we only displayed the accounts when all were retrieved)

Old UI 
https://github.com/user-attachments/assets/6752d858-589c-426a-9816-471ed8aa6772

New UI (with low network to see the accounts arriving one by one)

https://github.com/user-attachments/assets/40893889-f855-44c2-89bb-e2bdd461c7fa

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/login/components/OnboardingScreen.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/login/components/OnboardingScreen.kt
@@ -39,7 +39,8 @@ import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.unit.dp
 import com.infomaniak.core.compose.basics.ButtonType
 import com.infomaniak.core.compose.basics.Typography
-import com.infomaniak.core.crossapplogin.back.BaseCrossAppLoginViewModel.Companion.filterSelectedAccounts
+import com.infomaniak.core.crossapplogin.back.BaseCrossAppLoginViewModel.AccountsCheckingState
+import com.infomaniak.core.crossapplogin.back.BaseCrossAppLoginViewModel.AccountsCheckingStatus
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/ui/login/components/OnboardingScreen.kt
```diff
@@ -77,12 +78,12 @@ fun OnboardingScreen(
...
                     .consumeWindowInsets(paddingValues),
                 pagerState = pagerState,
-                accounts = accounts,
+                accountsCheckingState = accountsCheckingState,
                 skippedIds = skippedIds,
                 isLoginButtonLoading = isLoginButtonLoading,
                 isSignUpButtonLoading = isSignUpButtonLoading,
                 onLogin = { onLoginRequest(emptyList()) },
-                onContinueWithSelectedAccounts = { onLoginRequest(accounts().filterSelectedAccounts(skippedIds())) },
+                onContinueWithSelectedAccounts = { selectedAccounts -> onLoginRequest(selectedAccounts) },
```
LouisCAD, line 86: Looked into the code, and it was effectively redundant. Nice catch

LouisCAD (approved):

LouisCAD (approved):
Couldn't test, but the code changes are okay to me.

Added a dependency to a small PR to account for the changes in core that I submitted in https://github.com/Infomaniak/android-core/pull/598

PR #1888, opened 2025-11-26
fix: On chromeOS document provider queryChildDocuments now synchronized

baillyjamy:
Currently, there is a bug in the Chrome OS file explorer.
When the contentResolver is notified asynchronously, the file explorer does not necessarily refresh the view and does not request the updated file list again.
The file list is therefore only loaded synchronously on Chrome OS.


LouisCAD (approved):
We need to revisit this as we're kinda breaking structured concurrency in passing a `Job` to `launch`, which replaces the one of our `cloudScope`.

Good enough for now though.

PR #1891, opened 2025-12-02
fix: Download favorites files as offline

sirambd:
Files we never opened in local couldn't be download as offline because we created a fake path, then this local file was deleted because this fake path wasn't supported.
The path for folder opened from Favorite fragment was also wrong because of non tailrec path generation function

We now always fetch the remote file's path when we don't have a correct local path

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/bottomSheetDialogs/FileInfoActionsBottomSheetDialog.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/bottomSheetDialogs/FileInfoActionsBottomSheetDialog.kt
@@ -267,9 +267,9 @@ class FileInfoActionsBottomSheetDialog : EdgeToEdgeBottomSheetDialog(), FileInfo
         }
     }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/drive/utils/DownloadOfflineFileManager.kt
```diff
@@ -125,12 +120,26 @@ class DownloadOfflineFileManager(
...
+    }
+
     private fun getFileFromRemote(
         context: Context,
         fileId: Int,
         userDrive: UserDrive = UserDrive(),
-        onFileDownloaded: (downloadedFile: File) -> Unit
-    ) {
+    ): File {
         val fileDetails = ApiRepository.getFileDetails(File(id = fileId, driveId = userDrive.driveId))
```
FabianDevel: Shouldn't this use a withContext(Dispatchers.io) ?

FabianDevel (changes requested):

Issue #1909, opened 2025-12-23
F-Droid build failed


linsui:
https://github.com/Infomaniak/android-kDrive/pull/1900 add the InAppUpdate and InAppReview. InAppReview depends on google play lib which is not FOSS. InAppUpdate seems to be fine but we require the in app update check to be opt-in and it's useless for F-Droid users. Could you please exclude them from fdroid flavor? Thanks!

baillyjamy:
Hello, thank you for the report.
Sorry, I forgot to make sure that the F-Droid variant didn't implement Google services. 
It's now fixed. It will be available on the next release

linsui:
Thanks!

PR #1911, opened 2025-12-23
fix: Bump core to fix fdroid variant

baillyjamy:
Bump of core to implement fix from https://github.com/Infomaniak/android-core/pull/647


baillyjamy:
Fixes https://github.com/Infomaniak/android-kDrive/issues/1909

PR #1991, opened 2026-03-27
feat: Retrieve file's dates for provided uri and add them to fake File

benjaminVadon:
depends on : https://github.com/Infomaniak/android-core/pull/734

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
@@ -50,8 +50,6 @@ import com.infomaniak.drive.utils.shareFile
 import com.infomaniak.drive.utils.toggleSystemBar
 import com.infomaniak.drive.views.FileInfoActionsView.OnItemClickListener
 import kotlinx.coroutines.CoroutineStart
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.invoke
 import kotlinx.coroutines.launch
 import java.util.Date
 
@@ -136,9 +134,9 @@ class PreviewPDFActivity : AppCompatActivity(), OnItemClickListener {
         )
     }
 
-    private suspend fun Uri?.getDates(): Pair<Date?, Date> = Dispatchers.IO {
-        this@getDates?.retrieveAndUse { getFileDatesWithFallback() }
-    } ?: Pair(Date(), Date())
+    private suspend fun Uri?.getDates(): Pair<Date?, Date> {
+        return this@getDates?.retrieveAndUse { getFileDatesWithFallback() } ?: Pair(Date(), Date())
+    }
 
     private fun setupNavController(): NavController {
         return navHostFragment.navController.apply {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
```diff
@@ -117,15 +121,29 @@ class PreviewPDFActivity : AppCompatActivity(), OnItemClickListener {
...
+    private fun getDates(externalFileUri: Uri?): Pair<Date?, Date> {
+        return externalFileUri?.let {
+            contentResolver.query(externalFileUri, null, null, null, null)?.use { cursor ->
+                cursor.takeIf { cursor.moveToFirst() }?.getFileDatesWithFallback()
+            }
+        } ?: Pair(Date(), Date())
+    }
+
+    private val Date.wholeSeconds: Long
+        get() = time.milliseconds.inWholeSeconds
```
tevincent: Was that really necessary to create a `private val` with a getter instead of just an extension function?

tevincent (changes requested):

app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
LunarX: If the idea is to recreate the same signature as the Duration one, I can get behind the extension value. But in this case, maybe we'd want to stick with the original name `inWholeSeconds` as well

benjaminVadon: Yes it was the idea @LunarX . 
At first, I wanna put it in a Date extension file, but it don't exist in core so I let it here in private.
What is the best practice in your opinion, migrate to a core Date or let in private here (and thus I will change it to a getter function) ?

benjaminVadon: After some thought, I've migrated to a DateExt in Drive. We can move it to core if you think it's better

benjaminVadon: Finally moved to core, as I've a modification to do there

benjaminVadon: So added a dependency to https://github.com/Infomaniak/android-core/pull/734

tevincent (approved):

app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
```diff
@@ -116,16 +121,23 @@ class PreviewPDFActivity : AppCompatActivity(), OnItemClickListener {
...
             id = ROOT_ID,
             extensionType = ExtensionType.PDF.value,
             type = "",
+            createdAt = createdAt.inWholeSeconds(),
+            lastModifiedAt = updatedAt.inWholeSeconds,
         )
     }
 
+    private suspend fun Uri?.getDates(): Pair<Date?, Date> {
+        return this@getDates?.retrieveAndUse { getFileDatesWithFallback() } ?: Pair(Date(), Date())
```
FabianDevel, line 138: Shouldn't you wrap this in a dispatchers.IO ?

FabianDevel (changes requested):

FabianDevel (approved):

PR #1998, opened 2026-04-14
feat: Async auto deletion

benjaminVadon:
Better review by commit

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
@@ -1,6 +1,6 @@
 /*
  * Infomaniak kDrive - Android
- * Copyright (C) 2022-2025 Infomaniak Network SA
+ * Copyright (C) 2022-2026 Infomaniak Network SA
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -458,7 +458,7 @@ class MainActivity : BaseActivity() {
             }
         }
 
-        fun onConfirmation(filesUploadedRecently: List<UploadFile>, filesUriToDelete: List<Uri>) {
+        fun onConfirmation(filesUriToDelete: List<Uri>) {
             if (SDK_INT >= 30) {
                 lifecycleScope.launch {
                     pendingFilesUrisQueue.clear()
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -469,24 +463,38 @@ class MainActivity : BaseActivity() {
                 mainViewModel.deleteSynchronizedFilesOnDevice(filesUriToDelete)
             }
         }
+        lifecycleScope.launch(Dispatchers.IO) {
```
FabianDevel, line 466: Add a blank line above

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -452,41 +452,49 @@ class MainActivity : BaseActivity() {
...
-
-        deleteLocalMediaRequestDialog = Utils.createConfirmation(
-            context = this,
-            title = getString(R.string.modalDeletePhotosTitle),
-            message = getString(R.string.modalDeletePhotosNumericDescription, filesUploadedRecently.size),
-            buttonText = getString(R.string.buttonDelete),
-            isDeletion = true,
-            onConfirmation = { onConfirmation(filesUploadedRecently, filesUriToDelete) }
-        )
+    private suspend fun retrieveFilesUriToDelete(): List<Uri>? {
```
FabianDevel, line 482: I'd rather wrap this whole function with a `withContext(Dispatchers.IO)`, and launch a coroutine on the default dispatchers, that doing the other way with a withContext(Main).

I think this would enforce more the fact that the operations in this function need IO, and managing this function-scoped level will reduce possible errors

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
FabianDevel: You can do this change in the #1999

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
benjaminVadon: It was cleaned in next PR

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
benjaminVadon: Yes it will be clearer, I will do in #1999

FabianDevel (approved):

PR #1999, opened 2026-04-14
fix: Fix delete pictures

benjaminVadon:
depends on : https://github.com/Infomaniak/android-kDrive/pull/1998

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
@@ -445,10 +445,11 @@ class MainActivity : BaseActivity() {
     }
 
     private fun launchNextDeleteRequest() {
-        val filesUris = pendingFilesUrisQueue.firstOrNull() ?: return
         if (SDK_INT >= 30) {
-            val deletionRequest = MediaStore.createDeleteRequest(contentResolver, filesUris)
-            filesDeletionResult.launch(IntentSenderRequest.Builder(deletionRequest.intentSender).build())
+            pendingFilesUrisQueue.firstOrNull()
+                ?.let { runCatching { MediaStore.createDeleteRequest(contentResolver, it) }.getOrNull() }
+                ?.let { IntentSenderRequest.Builder(it.intentSender).build() }
+                ?.let(filesDeletionResult::launch)
         }
     }
 
--- a/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
@@ -481,7 +481,7 @@ class MainActivity : BaseActivity() {
             }
     }
 
-    private suspend fun showDeleteFileConfirmation(uris: List<Uri>) = withContext(Dispatchers.Main) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -491,9 +460,50 @@ class MainActivity : BaseActivity() {
...
+            .partition(Uri::doesFileExist)
+
+        UploadFile.deleteAllFromUris(filesAlreadyDeleted)
+
+        return filesToDelete
+    }
+
+    private fun Uri.isFileOrDocument(): Boolean {
+        return (scheme == ContentResolver.SCHEME_FILE || DocumentsContract.isDocumentUri(this@MainActivity, this))
+            .also { if (it) SentryLog.wtf("MainActivity", "That should never happen and if that happens") }
```
FabianDevel: "AND IF THAT HAPPENS... I'LL DO SOMETHING OK ??" 😂

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
```diff
@@ -491,9 +460,50 @@ class MainActivity : BaseActivity() {
...
+            message = getString(R.string.modalDeletePhotosNumericDescription, uris.size),
+            buttonText = getString(R.string.buttonDelete),
+            isDeletion = true,
+            onConfirmation = { onDeleteFileConfirmation(uris) }
+        )
+    }
+
+    fun onDeleteFileConfirmation(filesUriToDelete: List<Uri>) {
+        if (SDK_INT >= 30) {
+            lifecycleScope.launch {
```
FabianDevel: Is this launch necessary ? 
I checked createDeleteRequest, and it seems to just compute a PendingIntent

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
benjaminVadon: Yes it was a change done by Joris at first, I was not sure how to manage too 😅

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
benjaminVadon: Maybe for the chunk method, I don't know I didn't think about it. Or to prevent popup the system dialog if the activity lifecycle is stopped

benjaminVadon: Apparently it was to "chore: Move to background task" a7f5176e454495d4e667c2a1638fa9a09bdf699d but in fact not

app/src/main/java/com/infomaniak/drive/ui/MainActivity.kt
FabianDevel: The first version was "we must keep this if"

benjaminVadon: Seen together, the goal was to add a message to the dev to check if the sentry appears or not, and adapt the code then

FabianDevel (approved):

PR #2017, opened 2026-05-01
fix: On Xiaomi phone, lastModified is wrong so we fallback on the current time

tevincent:


app/src/main/java/com/infomaniak/drive/data/models/UploadFile.kt
```diff
@@ -154,6 +154,12 @@ open class UploadFile(
         }
     }
 
+    fun getLastModified(): Long {
```
benjaminVadon, line 157: Rather call "buildLastModified" or something similar to prevent misusage on getting lastModified date of an uploadFIle

benjaminVadon: And maybe private if possible with the usage you did of it

baillyjamy (approved):

PR #2021, opened 2026-05-13
fix: Support SAF children relation

sirambd:
Depends on #2022

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
+++ b/app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
@@ -890,7 +890,7 @@ class CloudStorageProvider : DocumentsProvider() {
         val fileFolderId = getFileIdFromDocumentId(documentId)
         val isMySharesRoot = isSharedUri(documentId)
                 && fileFolderId == Utils.ROOT_ID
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/drive/data/documentprovider/CloudStorageProvider.kt
```diff
@@ -277,6 +278,34 @@ class CloudStorageProvider : DocumentsProvider() {
...
+        documentType: CloudDocumentType.FileOrFolder,
+    ): Boolean = when {
+        parentDocumentType.userDrive == documentType.userDrive -> {
+            FileController.getRealmInstance(documentType.userDrive).use { realm ->
+                val folderProxy = FileController.getFileProxyById(documentType.fileId, customRealm = realm)
+                folderProxy?.parentId == parentDocumentType.fileId
+            }
+        }
+        else -> false
+    }
```
LunarX, line 315: This only checks for direct paternity of the file (i.e. one single level) but not multiple levels. Like if we checked `isChildDocument()` between `parent1` and `file` in this situation in `parent1/parent2/parent3/file`

LunarX (changes requested):

LunarX (approved):

PR #2048, opened 2026-06-30
fix: Wrong folder issue

NicolasBourdin88:


app/src/main/java/com/infomaniak/drive/ui/addFiles/AddFileBottomSheetDialog.kt
```diff
@@ -147,6 +142,31 @@ class AddFileBottomSheetDialog : EdgeToEdgeBottomSheetDialog() {
...
 
+    private fun observeCurrentFolder() {
+        val currentFolderMerged = MediatorLiveData<File?>().apply {
+            addSource(mainViewModel.currentFolderOpenAddFileBottom) { folderFromBottom ->
+                value = folderFromBottom ?: mainViewModel.currentFolder.value
+            }
+            addSource(mainViewModel.currentFolder) { folder ->
+                if (mainViewModel.currentFolderOpenAddFileBottom.value == null) value = folder
+            }
+        }
```
FabianDevel, line 153: This should be in the viewmodel

app/src/main/java/com/infomaniak/drive/ui/addFiles/AddFileBottomSheetDialog.kt
```diff
@@ -147,6 +142,31 @@ class AddFileBottomSheetDialog : EdgeToEdgeBottomSheetDialog() {
...
+    }
+
+    private fun updateCurrentFolder(file: File?) {
+        if (file == null) {
+            findNavController().popBackStack()
+            return
+        }
+
+        currentFolderFile = file
+        viewLifecycleOwner.lifecycleScope.launch(start = CoroutineStart.UNDISPATCHED) {
```
FabianDevel, line 165: you're blocking the UI thread here
It looks like you need suspending because of a realm read for the categories
I guess we should better use the withContext there.

app/src/main/java/com/infomaniak/drive/ui/addFiles/AddFileBottomSheetDialog.kt
```diff
@@ -285,6 +321,11 @@ class AddFileBottomSheetDialog : EdgeToEdgeBottomSheetDialog() {
         trackNewElementEvent(trackerName + trackerSource)
     }
 
+    private fun showErrorCurrentFolder() {
```
FabianDevel, line 324:
```suggestion
    private fun showErrorAndPopBack() {
```

FabianDevel (changes requested):
I feel like it should be better to have the folderId by navigation and then fetch it from realm.

Or at least think about simplifying this, because the currentFolderOpenAddFileBottom should probably be removed.

FabianDevel:
This doesn't fix the problem it adressed. We finally found the root cause and it will be fixed in another PR

PR #2055, opened 2026-07-17
fix: Fix sharedWithMe files sometimes being wrongly queried int the DB

FabianDevel:
As SharedWithMe files are all in a single DB but originate from multiple drives, several files can have the same fileId (because it's their id on their parent's drive). 

We were querying the first matching id from the DB, thus we could have the wrong file.

With this new getFileByUidOrId, we now query the file by their uid in the `sharedWithMeFragment` to avoid this problem.


NicolasBourdin88 (approved):

NicolasBourdin88 (changes requested):
Need to test

NicolasBourdin88 (approved):

sirambd (approved):

PR #2056, opened 2026-07-22
fix: PreviewPDFActivity was not handling edge to edge properly

tevincent:
|AndroidVersion | Before    | After |
| --------------- | -------- |------ |
Android < 15 |  Screenshot_20260722_110717 |  Screenshot_20260722_110822  |
Android > 15 | Screenshot_20260722_110555 | Screenshot_20260722_110931   |

### DIFF
--- a/app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
+++ b/app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
@@ -20,18 +20,19 @@ package com.infomaniak.drive.ui.fileList.preview
 import android.net.Uri
 import android.os.Bundle
 import android.view.View
-import androidx.appcompat.app.AppCompatActivity
 import androidx.core.net.toUri
 import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavController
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
```diff
@@ -79,7 +80,7 @@ class PreviewPDFActivity : AppCompatActivity(), OnItemClickListener {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-
+        window.lightStatusBar(false)
```
FabianDevel, line 83: We use `window.lightStatusBar(!isNightModeEnabled())`

Maybe we should here too

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/drive/ui/fileList/preview/PreviewPDFActivity.kt
tevincent: No, it's in the preview. it's always on a dark background.

FabianDevel (approved):

PR #2063, opened 2026-08-05
feat: Sort gallery by period

aymericmariaux:


app/src/main/java/com/infomaniak/drive/ui/menu/GalleryViewModel.kt
```diff
@@ -31,6 +31,8 @@ import kotlinx.coroutines.launch
 
 class GalleryViewModel : ViewModel() {
 
+    var period: GalleryPeriod = GalleryPeriod.MONTH
```
tevincent: Instead of having a `var`, you could have a flow instead and react to it in the fragment.

app/src/main/java/com/infomaniak/drive/ui/menu/GalleryFragment.kt
```diff
@@ -270,6 +270,15 @@ class GalleryFragment : MultiSelectFragment(
         }
     }
 
+    val currentPeriod get() = galleryViewModel.period
```
tevincent: Why do you need that ?

tevincent (changes requested):

tevincent (approved):

